PR #1, opened 2023-01-20
Initial monitoring-plugin code

kyoto:


src/components/console/graphs/graph-empty.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+
+export const GraphEmpty: React.FC<GraphEmptyProps> = ({ height = 180, loading = false }) => {
+  const { t } = useTranslation();
+
+  return (
+    <div
+      style={{
```
jgbernalp, line 9: Should we put this style in a separate file?

package.json
```diff
@@ -0,0 +1,97 @@
+{
+  "name": "monitoring-console-plugin",
+  "version": "0.0.1",
+  "private": true,
+  "repository": "git@github.com:kyoto/monitoring-console-plugin.git",
```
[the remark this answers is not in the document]
kyoto: Oops. Thank you.

src/components/console/graphs/graph-empty.tsx
kyoto: Everything under the `console/` directory is copied from the console codebase, so I'm not inclined to do any tidying up of that code here. Over time, everything under `console/` should be refactored anyway.

src/components/alerting.tsx
```diff
@@ -0,0 +1,2167 @@
...
+  BanIcon,
+  BellIcon,
+  BellSlashIcon,
+  HourglassHalfIcon,
+  OutlinedBellIcon,
+} from '@patternfly/react-icons';
+import { sortable } from '@patternfly/react-table';
+import classNames from 'classnames';
+import i18next from 'i18next';
+import * as _ from 'lodash-es';
```
jgbernalp, line 62: Should we refactor the usages to avoid importing the whole library?

kyoto: It may be worth doing that yes. However, lodash is used a lot in this file so I think it's better to make it a separate PR.

jgbernalp:
It would be nice to add a OWNERS file as this will be required by the CI.

src/components/alerting.tsx
kyoto: I added a task for this: https://issues.redhat.com/browse/OU-139

kyoto:
@jgbernalp Addressed your comments

PR #4, opened 2023-02-09
Add Helm charts for installing plugin

kyoto:
Helm charts are taken from github.com/openshift/console-plugin-template

### DIFF
--- a/charts/openshift-console-plugin/values.yaml
+++ b/charts/openshift-console-plugin/values.yaml
@@ -1,0 +1,31 @@
+---
+plugin:
+  name: "monitoring-plugin"
+  description: ""
+  image: ""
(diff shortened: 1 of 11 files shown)


charts/openshift-console-plugin/values.yaml
```diff
@@ -0,0 +1,31 @@
+---
+plugin:
+  name: ""
```
jgbernalp, line 3:
```suggestion
  name: "monitoring-plugin"
```

jgbernalp (commented):
just a nit, the rest LGTM

charts/openshift-console-plugin/values.yaml
kyoto: 👍

PR #13, opened 2023-02-20
Improve console extensions including flags

jgbernalp:
This PR adjusts the extensions to match the original ones that consider flags to hide or show the observe menu and submenus: https://github.com/openshift/console/blob/master/frontend/packages/console-app/console-extensions.json#L1261
It also adds i18n files to provide translations for the menu items, taken from: https://github.com/openshift/console/tree/master/frontend/packages/console-app/locales

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -9,42 +9,58 @@
   },
   {
     "type": "console.navigation/href",
+    "flags": {
+      "required": ["PROMETHEUS", "MONITORING", "CAN_GET_NS"]
+    },
     "properties": {
       "id": "alerting",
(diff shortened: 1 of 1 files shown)


kyoto:
Thanks @jgbernalp. The `locale/*/` files are no longer relevant, but the `console-extensions.json` changes are. Could you please remove the locale files?

kyoto (changes requested):
I see that the `monitoring` prefix was removed from each `id`, so I think it also needs to be removed from each `insertAfter` to match.

kyoto (approved):
/lgtm

PR #14, opened 2023-03-02
[WIP] Rework Dockerfile to not use npm

kyoto:


kyoto:
We will use the approach in https://github.com/openshift/monitoring-plugin/pull/15 instead

PR #28, opened 2023-04-26
Add translations and add standalone version for Cypress tests

kyoto:
Add English translations, but not other language translations for now since they are not needed for the current Cypress tests. Leave the i18n namespace as `public` for now so that we can fall back to translations available in the web console.


kyoto:
Fixed lint errors

PR #31, opened 2023-05-11
OU-158: Add bulk expire for silences

kyoto:
Adds the ability to bulk expire silences from the silences list page.

Adds a checkbox per row to the silences list for selecting silences as well as a checkbox for selecting / unselecting all of the silences.

If the list is currently filtered, the select all checkbox only selects the currently displayed silences, but unselecting all also unselects any silences that are currently hidden.

### DIFF
--- a/src/components/alerting.tsx
+++ b/src/components/alerting.tsx
@@ -31,6 +31,7 @@ import {
   Breadcrumb,
   BreadcrumbItem,
   Button,
+  Checkbox,
   CodeBlock,
   CodeBlockCode,
   Dropdown,
@@ -118,6 +119,11 @@ import {
   SilenceResource,
   silenceState,
 } from './utils';
+
+const SelectedSilencesContext = React.createContext({
+  selectedSilences: new Set(),
+  setSelectedSilences: undefined,
+});
 
 const ruleURL = (rule: Rule) => `${RuleResource.plural}/${_.get(rule, 'id')}`;
 
@@ -463,6 +469,7 @@ const Graph: React.FC<GraphProps> = ({
 };
 
 const tableSilenceClasses = [
+  'pf-c-table__action', // Checkbox
   'pf-u-w-50 pf-u-w-33-on-sm', // Name
   'pf-m-hidden pf-m-visible-on-sm', // Firing alerts
   '', // State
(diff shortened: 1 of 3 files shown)


src/components/alerting.tsx
```diff
@@ -1874,18 +1906,61 @@ const RulesPage_: React.FC<{}> = () => {
...
+
+  const [isInProgress, , setInProgress, setNotInProgress] = useBoolean(false);
+
+  const dispatch = useDispatch();
+
+  const { selectedSilences } = React.useContext(SelectedSilencesContext);
+
+  const onClick = () => {
+    setInProgress();
+    Promise.all(
```
jgbernalp, line 1936: should we use `allSettled` instead in the case some succeed and some fail?

src/components/alerting.tsx
```diff
@@ -1902,9 +1977,32 @@ const silenceStateOrder = (silence: Silence) => [
   _.get(silence, silenceState(silence) === SilenceStates.Pending ? 'startsAt' : 'endsAt'),
 ];
 
+const SelectAllCheckbox: React.FC<{ silences: Silence[] }> = ({ silences }) => {
+  const { selectedSilences, setSelectedSilences } = React.useContext(SelectedSilencesContext);
+
+  const activeSilences = _.filter(silences, (s) => silenceState(s) !== SilenceStates.Expired);
```
jgbernalp, line 1983: should we avoid using lodash references from new code so when we refactor it away is easier?

kyoto: My understanding was that we're not aiming to completely refactor away our use of Lodash, but just to refactor away places where it is used with very little benefit. (Arguably this use of `_.filter` fits that definition 😅). We should definitely reduce our use of Lodash, but not sure where the line should be drawn.

jgbernalp: I think we could agree on prefer to use native functions when possible, then we could use lodash for something that provided by default. WDYT?

kyoto: Sounds good 👍 
There are some places where there really is no value added by using Lodash, so let's start be refactoring those away.

src/components/alerting.tsx
kyoto: Follow up PR to change this: https://github.com/openshift/monitoring-plugin/pull/34

PR #36, opened 2023-05-17
Fix CSS not being compiled into dist/ directory

kyoto:
Don't rely on any global SASS variables being defined. There should be no effect on any visible styles except for the timestamp `font-size`, which is increased by this change form `13px` to `14px`.

Also removes some redundant CSS rules from the `monitoring-timestamp` CSS class.


src/components/_monitoring.scss
```diff
@@ -1,3 +1,7 @@
+$screen-sm-min: 768px;
+$screen-lg-min: 1200px;
+$pf-color-black-700: #26292d;
```
jgbernalp, line 3: wouldn't be better to use the actual var from Patternfly?

PR #53, opened 2023-06-27
OU-179: Fix the root cause of externalLabels not present on alerts

zhuje:
[OU-179](https://issues.redhat.com//browse/OU-179) Display externalLabels found in GET `/api/prometheus/api/v1/rules` response >> `rules.labels`.

**Related Issue:** 
https://issues.redhat.com/browse/OU-179

**AlertDetailsPage**
External labels are now listed alongside the other labels associated with an alert.
Screenshot 2023-06-27 at 11 37 38 AM

**Search Label** 
The external labels can be pre-populated in the 'label' search field. 
Search External Label

**Silences**
We're able to silence alerts associated with the external label. 
Create Silences for External Label

**To Test:** 
1. Attach an externalLabel to all time-series and alerts leaving the Prometheus instance that monitors core OpenShift Container Platform projects [[docs]](https://docs.openshift.com/container-platform/4.13/monitoring/configuring-the-monitoring-stack.html#attaching-additional-labels-to-your-time-series-and-alerts_configuring-the-monitoring-stack)
2. Go to OpenShift Console UI > Observe > Alerts > AlertsDetailPage should now display the externalLabels (this may take a few minutes to update).

### DIFF
--- a/src/components/utils.ts
+++ b/src/components/utils.ts
@@ -71,6 +71,11 @@ export const getAlertsAndRules = (
     return _.filter(g.rules, { type: 'alerting' }).map(addID);
   });
 
+  // Add external labels to all `rules[].alerts[].labels`
+  rules.forEach((rule) => {
+    rule.alerts.forEach((alert) => (alert.labels = { ...rule.labels, ...alert.labels }));
+  });
+
   // Add `rule` object to each alert
   const alerts = _.flatMap(rules, (rule) => rule.alerts.map((a) => ({ rule, ...a })));
 


zhuje:
We received a confirmation from Simon Pasquier the API response from `/api/prometheus/api/v1/rules` exhibits the expected behavior from Thanos/Prometheus on August 8, 2023 in the Internal Red Hat Slack > # forum-openshift-monitoring : 

> this is indeed the expected behavior:
https://github.com/thanos-io/thanos/blob/c492ce9e41fdc5b235860c89ae8f81ce86bc090b/pkg/rules/prometheus.go#L45-L46
https://github.com/thanos-io/thanos/blob/c492ce9e41fdc5b235860c89ae8f81ce86bc090b/pkg/rules/prometheus.go#L56-L63
my guess is that since the rules carry the external labels and to avoid verbosity, there's no need to repeat them into the alerts

src/components/alerting.tsx
```diff
@@ -726,6 +726,13 @@ const AlertsDetailsPage_: React.FC<AlertsDetailsPageProps> = ({ history, match }
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const labels: PrometheusLabels = React.useMemo(() => alert?.labels, [labelsMemoKey]);
 
+  const ruleLabelsMemoKey = JSON.stringify(rule?.labels);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const ruleLabels: PrometheusLabels = React.useMemo(() => rule?.labels, [ruleLabelsMemoKey]);
+
+  const combinedLabels = { ...labels, ...ruleLabels };
```
kyoto, line 733: This doesn't look right to me. If `labels` and `ruleLabels` contain the same key, this will take the value from `ruleLabels`, which seems wrong?

src/components/alerting.tsx
```diff
@@ -726,6 +726,13 @@ const AlertsDetailsPage_: React.FC<AlertsDetailsPageProps> = ({ history, match }
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const labels: PrometheusLabels = React.useMemo(() => alert?.labels, [labelsMemoKey]);
 
+  const ruleLabelsMemoKey = JSON.stringify(rule?.labels);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const ruleLabels: PrometheusLabels = React.useMemo(() => rule?.labels, [ruleLabelsMemoKey]);
```
kyoto, line 731: Unlike with `labels`, I don't think there's really a benefit in memoizing `ruleLabels`. Memoizing `labels` prevents unnecessary rerendering of several components including the `Graph`, which is quite slow to render. Memoizing `ruleLabels` doesn't look like it will prevent any rerenders or improve performance generally.

src/components/alerting.tsx
```diff
@@ -726,6 +726,13 @@ const AlertsDetailsPage_: React.FC<AlertsDetailsPageProps> = ({ history, match }
...
+
+  const combinedLabels = { ...labels, ...ruleLabels };
+  delete combinedLabels?.prometheus;
```
kyoto, line 734: What's the reason for deleting the `prometheus` label? If it should be deleted, let's add a comment explaining  why.

kyoto (changes requested):
I think there's more complexity here unfortunately.

We presumably should also be able to search for alerts by external label on the alerts list page.

When silencing an alert, should the external labels be included in the default label matchers that we auto-generate when the create silence page is opened?

For the `AlertMessage`s below, should they also be receiving the external labels in case those labels are used in generating the alert message / description?

src/components/alerting.tsx
zhuje: This code has been removed in latest commit.

src/components/alerting.tsx
zhuje: This code has been removed in latest commit.

src/components/utils.ts
```diff
@@ -74,6 +74,9 @@ export const getAlertsAndRules = (
   // Add `rule` object to each alert
   const alerts = _.flatMap(rules, (rule) => rule.alerts.map((a) => ({ rule, ...a })));
 
+  // `alert.rule.labels` contains external labels which we want to include with `alert.labels`.
+  alerts.map((alert) => (alert.labels = { ...alert.labels, ...alert.rule.labels }));
```
kyoto, line 78:
```suggestion
  alerts.map((alert) => (alert.labels = { ...alert.rule.labels, ...alert.labels }));
```

If the rule and the alert have a label with the same key, I think we want the alert's label to be preserved.

kyoto (changes requested):
Thanks @zhuje. I think this is probably the right approach so that the external labels are available throughout the UI.

src/components/utils.ts
zhuje: I see what you're saying now. Yes, that makes sense, I'll update this in the next commit.

kyoto:
I discovered one problem with this. On the rule details page, there is a list of currently firing alerts for the rule. With this change, the links in that list are broken. This happens because the labels form part of the URL and the URLs are generated using `rule.alerts[].labels`, but the alert details page uses the labels from `alerts[].labels` (they are assumed to be the same).

src/components/utils.ts
```diff
@@ -74,6 +74,9 @@ export const getAlertsAndRules = (
   // Add `rule` object to each alert
   const alerts = _.flatMap(rules, (rule) => rule.alerts.map((a) => ({ rule, ...a })));
 
+  // `alert.rule.labels` contains external labels which we want to include with `alert.labels`.
+  alerts.map((alert) => (alert.labels = { ...alert.rule.labels, ...alert.labels }));
```
kyoto, line 78:
```suggestion
  // Add external labels to all `rules[].alerts[].labels`
  rules.map(...);
  
  const alerts = _.flatMap(rules, (rule) => rule.alerts.map((a) => ({ rule, ...a })));
```

As mentioned in my other comment, I think we need to add the external labels not just to all the `alerts[].labels`, but also to all the `rules[].alerts[].labels`. So if we add them to `rules[].alerts[].labels` first, the existing `const alerts = _.flatMap(rules, (rule) => rule.alerts.map((a) => ({ rule, ...a })))` line should take care of copying them to `alerts[].labels`.

kyoto (changes requested):

src/components/utils.ts
zhuje: Thanks, good catch! I've updated this in the latest push.

src/components/utils.ts
```diff
@@ -71,6 +71,13 @@ export const getAlertsAndRules = (
     return _.filter(g.rules, { type: 'alerting' }).map(addID);
   });
 
+  // Add external labels to all `rules[].alerts[].labels`
+  rules.map((rule) => {
+    if (rule.alerts.length > 0) {
+      rule.alerts.map((alert) => (alert.labels = { ...rule.labels, ...alert.labels }));
+    }
+  });
```
jgbernalp, line 79: It seems `map` is not needed here:

```suggestion
  rules.forEach((rule) => {
    if (rule.alerts.length > 0) {
      rule.alerts.forEach((alert) => (alert.labels = { ...rule.labels, ...alert.labels }));
    }
  });
```

zhuje: Thanks! Committed and squashed in the latest commit.

src/components/utils.ts
```diff
@@ -71,6 +71,13 @@ export const getAlertsAndRules = (
     return _.filter(g.rules, { type: 'alerting' }).map(addID);
   });
 
+  // Add external labels to all `rules[].alerts[].labels`
+  rules.forEach((rule) => {
+    if (rule.alerts.length > 0) {
```
kyoto, line 76: I think the `if (rule.alerts.length > 0)` should be redundant since the `forEach` will do nothing for an empty array anyway?

juzhao: @zhuje do we need to update code based on comment from @kyoto ?

zhuje: Updated in the latest commit.

zhuje: @juzhao would you be able to test one more time, please?

jgbernalp:
/hold
As no merges should be done to main until 4.15 is open

src/components/utils.ts
juzhao: @zhuje tested again, if this PR is only for admin console only, no issue, but for developer console, there is one bug [OU-252](https://issues.redhat.com/browse/OU-252), maybe we have not implement this feature to developer console

kyoto: Thanks @juzhao. This PR just affects the admin console, so it is expected that this doesn't add external labels to the developer console. We plan to transition the dev console Alerting page UX to match the admin console, which will resolve this.

kyoto (approved):
/lgtm

PR #59, opened 2023-07-25
Fix the props types for `SilencesPage` and `AlertsPage`

kyoto:


src/components/alerting.tsx
```diff
@@ -1685,7 +1685,7 @@ const getAdditionalSources = <T extends Alert | Rule>(
   return [];
 };
 
-const AlertsPage_: React.FC<Alerts> = () => {
+const AlertsPage_: React.FC<{}> = () => {
```
jgbernalp, line 1688:
```suggestion
const AlertsPage_: React.FC = () => {
```

src/components/alerting.tsx
```diff
@@ -2090,7 +2090,7 @@ const SelectAllCheckbox: React.FC<{ silences: Silence[] }> = ({ silences }) => {
   );
 };
 
-const SilencesPage_: React.FC<Silences> = () => {
+const SilencesPage_: React.FC<{}> = () => {
```
jgbernalp, line 2093:
```suggestion
const SilencesPage_: React.FC = () => {
```

jgbernalp (approved):

PR #65, opened 2023-08-03
OCPBUGS-15323: Use datasource defined in dashboard definition broken by switch to web console monitoring plugin

zhuje:
**Related Issue**: 
[OCPBUGS-15323](https://issues.redhat.com/browse/OCPBUGS-15323)

**Summary**: 
Fix custom dashboard logic to await extension resolution before using the extension. This is needed because of the update to use `useResolvedExtensions()` instead of `useExtension()` from the `openshift-console/dynamic-plugin-sdk`.


Tai-RedHat:
tested with cluster-bot, dashboard page works fine

PR #72, opened 2023-08-21
targets: filter by target

stevekuznetsov:
Previously, a user could only filter by the endpoint or namespace. Endpoints are generally the same between components and not useful for filtering, as well as being opaque. Namespaces are useful but not precise. Filtering by the target name (PodMonitor or ServiceMonitor) will be more useful for narrowing down the list.


kyoto:
/hold

As discussed on Slack, this is unfortunately more complicated because the monitors data is loaded asynchronously. It makes several requests to get the monitors data, which and can take a while. Allowing filtering by data that has not yet loaded likely leads to confusing UX 🙁

stevekuznetsov:
@kyoto can you think of any other middle ground here? I hope my commit message (and personal experience with this page) show how the filtering UX is almost not useful right now :|

kyoto:
Created Jira ticket so that we don't forget this feature request: https://issues.redhat.com/browse/OBSDA-531

PR #74, opened 2023-09-05
OU-261: Fix "Overwriting current silence" info alert to have padding

kyoto:
| Before | After|
| - | - |
| before | after |

FYI @fkargbo


fkargbo:
Looks good! 👍

PR #95, opened 2024-01-18
OU-310: Add an extension point for alerts, to be able to add an inspect link

zhuje:
**Description** 
This PR creates an extension point in the kebab menu on the alerts page. 

**Screenshots**
Screenshot 2024-01-18 at 3 48 03 PM
_Figure 1._ The screenshot above shows new kebab items representing actions received from the mocked extension.

### DIFF
--- a/src/components/alerting.tsx
+++ b/src/components/alerting.tsx
@@ -450,7 +450,12 @@ const SourceHelp: React.FC = () => {
 
 type ActionWithHref = Omit<Action, 'cta'> & { cta: { href: string; external?: boolean } };
 
+type ActionWithCallBack = Omit<Action, 'cta'> & { cta: () => void };
+
 const isActionWithHref = (action: Action): action is ActionWithHref => 'href' in action.cta;
+
+const isActionWithCallback = (action: Action): action is ActionWithCallBack =>
+  typeof action.cta === 'function';
 
 const queryBrowserURL = (query: string, namespace: string) =>
   namespace
@@ -1618,6 +1623,26 @@ const AlertTableRow_: React.FC<AlertTableRowProps> = ({ history, obj }) => {
     );
   }
 
+  const getDropdownItemsWithExtension = (actions: Action[]) => {
+    const extensionDropdownItems = [];
+    actions.forEach((action) => {
+      if (isActionWithHref(action)) {
+        extensionDropdownItems.push(
+          <DropdownItem key={action.id} href={action.cta.href}>
+            {action.label}
+          </DropdownItem>,
+        );
+      } else if (isActionWithCallback(action)) {
+        extensionDropdownItems.push(
+          <DropdownItem key={action.id} onClick={action.cta}>
+            {action.label}
+          </DropdownItem>,
+        );
+      }
+    });
+    return dropdownItems.concat(extensionDropdownItems);
+  };
+
   return (
     <>
       <td className={tableAlertClasses[0]} title={title}>
@@ -1644,7 +1669,15 @@ const AlertTableRow_: React.FC<AlertTableRowProps> = ({ history, obj }) => {
         {alertSource(obj) === AlertSource.User ? t('User') : t('Platform')}
       </td>
       <td className={tableAlertClasses[4]} title={title}>
-        <KebabDropdown dropdownItems={dropdownItems} />
+        <ActionServiceProvider context={{ 'monitoring-alert-list-item': { alert } }}>
+          {({ actions, loaded }) => {
+            if (loaded && actions.length > 0) {
+              return <KebabDropdown dropdownItems={getDropdownItemsWithExtension(actions)} />;
+            } else {
+              return <KebabDropdown dropdownItems={dropdownItems} />;
+            }
+          }}
+        </ActionServiceProvider>
       </td>
     </>
   );


console-extensions.json
```diff
@@ -78,5 +78,12 @@
       "section": "observe",
       "insertAfter": "dashboards"
     }
+  },
+	{
```
kyoto, line 82: Nit: Looks like a tab character instead of 2 space characters

kyoto:
This extension could be used by any other plugin, so I think it's better to remove "lightspeed" from the various names. Instead we can use names that reflect what the actions actually do like `useAlertActionsExtension`.

src/components/alerting.tsx
```diff
@@ -1618,6 +1618,25 @@ const AlertTableRow_: React.FC<AlertTableRowProps> = ({ history, obj }) => {
     );
   }
 
+  const handleLightSpeedActions = (actions: Action[]) => {
```
kyoto, line 1621: If this function got called twice, all the dropdown actions would be doubled. It feels safer to change this to a "getExtensionDropdownItems" style function that just returns the `DropdownItem`s from any extensions.

kyoto:
Thanks @zhuje! Couple of small suggestions, but generally looks great.

console-extensions.json
```diff
@@ -78,5 +78,12 @@
...
+    "type": "console.action/provider",
+    "properties": {
+      "contextId": "mock-lightspeed",
```
jgbernalp, line 85: I think this is only a mock for test, but it shouldn't be included in the code

package.json
```diff
@@ -100,7 +100,8 @@
     "displayName": "OpenShift console monitoring plugin",
     "description": "This plugin adds the monitoring UI to the OpenShift web console",
     "exposedModules": {
-      "MonitoringUI": "./components/alerting"
+      "MonitoringUI": "./components/alerting",
+      "MockLightspeed": "./components/mock-lightspeed"
```
jgbernalp, line 104:
```suggestion
```

src/components/alerting.tsx
```diff
@@ -1644,7 +1663,15 @@ const AlertTableRow_: React.FC<AlertTableRowProps> = ({ history, obj }) => {
         {alertSource(obj) === AlertSource.User ? t('User') : t('Platform')}
       </td>
       <td className={tableAlertClasses[4]} title={title}>
-        <KebabDropdown dropdownItems={dropdownItems} />
+        {/* TODO: replace 'mock-lightspeed' with actual light speed contextId */}
+        <ActionServiceProvider context={{ 'mock-lightspeed': { alert } }}>
```
jgbernalp, line 1667:
```suggestion
        <ActionServiceProvider context={{ 'monitoring-alert-list-item': { alert } }}>
```

src/components/mock-lightspeed.ts
```diff
@@ -0,0 +1,26 @@
+import { Action, ExtensionHook } from '@openshift-console/dynamic-plugin-sdk';
```
jgbernalp, line 1: we can remove this as this will be implemented by other plugins

jgbernalp (changes requested):

console-extensions.json
zhuje: Removed mocked extension in latest commit.

console-extensions.json
zhuje: Yes, it is only a mock for testing. It is removed in the latest commit.

package.json
zhuje: Mock extension used for testing. Removed in latest commit.

src/components/alerting.tsx
zhuje: Updated in the latest commit.

src/components/alerting.tsx
zhuje: Updated in the latest commit.

src/components/mock-lightspeed.ts
zhuje: Yes, this mock extension is for testing and is removed in the latest commit.

src/components/alerting.tsx
```diff
@@ -1618,6 +1618,26 @@ const AlertTableRow_: React.FC<AlertTableRowProps> = ({ history, obj }) => {
...
+  const getExtensionDropdownItems = (actions: Action[]) => {
+    const extensionDropdownItems = actions.map((action) => {
+      if (isActionWithHref(action)) {
+        return (
+          <DropdownItem key={action.id} href={action.cta.href}>
+            {action.label}
+          </DropdownItem>
+        );
+      } else {
+        const onClick = action.cta as () => void;
```
jgbernalp, line 1630: Do we need this cast?, at this point `action.cta` can only be a function

zhuje: It was giving me an error if I didn't cast it as a function...

src/components/alerting.tsx
```diff
@@ -1618,6 +1618,26 @@ const AlertTableRow_: React.FC<AlertTableRowProps> = ({ history, obj }) => {
     );
   }
 
+  const getExtensionDropdownItems = (actions: Action[]) => {
```
kyoto, line 1621:
```suggestion
  const getDropdownItemsWithExtension = (actions: Action[]) => {
```

Since we're actually returning all of the dropdown items.

Alternatively, could just return `extensionDropdownItems` and do the `concat` outside this function.

src/components/alerting.tsx
zhuje: Updated as discussed (off thread) with a type guard to check if action is a function. 
```
const isActionWithCallback = (action: Action): action is ActionWithCallBack =>
  typeof action.cta === 'function';
```

src/components/alerting.tsx
zhuje: That makes sense. Updated the function name to `getDropdownItemsWithExtension`.

kyoto (approved):
LGTM if @jgbernalp is happy that his comments have been addressed

PR #98, opened 2024-01-24
OU-236: Add CSV export option to dashboards and metrics tables

zhuje:
Add the ability to export data to a .csv file.

### DIFF
--- a/src/components/metrics.tsx
+++ b/src/components/metrics.tsx
@@ -223,6 +223,14 @@ const QueryKebab: React.FC<{ index: number }> = ({ index }) => {
     observe.getIn(['queryBrowser', 'queries', index, 'isEnabled']),
   );
 
+  const query = useSelector(({ observe }: RootState) =>
+    observe.getIn(['queryBrowser', 'queries', index, 'query']),
+  );
+
+  const queryTableData = useSelector(({ observe }: RootState) =>
+    observe.getIn(['queryBrowser', 'queries', index, 'queryTableData']),
+  );
+
   const dispatch = useDispatch();
 
   const toggleIsEnabled = React.useCallback(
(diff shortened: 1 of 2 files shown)


src/components/metrics.tsx
```diff
@@ -452,6 +479,21 @@ export const QueryTable: React.FC<QueryTableProps> = ({ index, namespace }) => {
...
+            variant="link"
+            isInline
+            icon={<ExternalLinkSquareAltIcon />}
+            iconPosition="right"
+            className="query-browser__table-export-link"
+          >
+            <CSVLink
+              data={csvData}
+              filename={`OpenShift_Metrics_QueryResultsTable_${query.replace(/\s/g, '')}.csv`}
+              enclosingCharacter={``}
```
kyoto, line 492:
```suggestion
              enclosingCharacter=""
```

src/components/metrics.tsx
```diff
@@ -436,6 +439,30 @@ export const QueryTable: React.FC<QueryTableProps> = ({ index, namespace }) => {
...
+  const getCsvRows = () => {
+    const csvRows = [];
+    for (let i = 0; i < rows.length; i++) {
+      const row = [...rows[i]].slice(1);
+      csvRows.push(row);
+    }
+    return csvRows;
+  };
+
+  const csvData = [getCsvColumns(), ...getCsvRows()];
```
kyoto, line 464: This is doing the work of preparing the CSV data for download every time the table is rendered. It would be much more efficient if we only generated the CSV data when the download button is clicked and it looks like `react-csv` does support that approach. (We've had trouble with performance on this Metrics page in the past and have worked hard to make it more efficient.)

src/components/metrics.tsx
```diff
@@ -436,6 +439,30 @@ export const QueryTable: React.FC<QueryTableProps> = ({ index, namespace }) => {
...
+  };
+
+  const getCsvRows = () => {
+    const csvRows = [];
+    for (let i = 0; i < rows.length; i++) {
+      const row = [...rows[i]].slice(1);
+      csvRows.push(row);
+    }
+    return csvRows;
+  };
```
kyoto, line 462: I guess this would be a little nicer if we use `map` instead of a `for` loop.

Also, I think `[...rows[i]].slice(1)` is equivalent to just `rows[i].slice(1)` since both `...` and `slice` create [shallow copies](https://developer.mozilla.org/en-US/docs/Glossary/Shallow_copy) of the array.

So we'd end up with just `const getCsvRows = () => rows.map((row) => row.slice(1));`

src/components/metrics.tsx
```diff
@@ -436,6 +439,30 @@ export const QueryTable: React.FC<QueryTableProps> = ({ index, namespace }) => {
...
+    for (let i = 1; i < columns.length; i++) {
+      const column = columns[i];
+      if (typeof column?.title === 'string') {
+        csvColumns.push(column.title);
+      } else if (column.title?.props?.children) {
+        csvColumns.push(column.title.props.children);
+      }
+    }
+    return csvColumns;
+  };
```
kyoto, line 453: If we don't find the title in either `column.title` or `column.title?.props?.children`, nothing will be pushed onto `csvColumns`, which will cause the titles to get out of sync with the data.

In other words, I think we need
```
if (typeof column?.title === 'string') {
  csvColumns.push(column.title);
} else if (column.title?.props?.children) {
  csvColumns.push(column.title.props.children);
} else {
  csvColumns.push('');
}
```

Also, I think this is a good case for using a `map`. Especially since using `map` will naturally guarantee that we get one output element for every input element.

So something like below (not tested!)
```
const getCsvColumns = () => {
  return columns.slice(1).map((column) => {
    if (typeof column?.title === 'string') {
      return column.title;
    } else if (column.title?.props?.children) {
      return column.title.props.children;
    } else {
      return '';
    }
  });
};
```

kyoto (changes requested):

src/components/metrics.tsx
```diff
@@ -71,6 +71,9 @@ import { colors, Error, QueryBrowser } from './query-browser';
 import TablePagination from './table-pagination';
 import { PrometheusAPIError, RootState } from './types';
 
+import { CSVLink } from 'react-csv';
+import ExternalLinkSquareAltIcon from '@patternfly/react-icons/dist/esm/icons/external-link-square-alt-icon';
```
kyoto, line 75: Shouldn't we be using the "Download" or "File Download' icon rather than the external link icon? (See https://preview.patternfly.org/icons/).

FYI @fkargbo

jgbernalp: Instead of this icon. The option should be included as part of the kebab menu as proposed on the mockups

fkargbo: Good catch @kyoto , that is the "external link" icon. PatternFly has an icon for "Download" and for "Export". We should probably align the icon that matches the text as per the Option 1 diagram if we're handling multiple file formats, i.e. CSV, PNG, PDF, etc... 

Alternatively, we could use just the text as shown in Option 3 if we're only exporting a single file format. Here's a link to the Google Design doc. https://docs.google.com/document/d/1OgU41FCRO6PGisZU66ky00KYrL9HhTObILlPQkHHonI/edit

Let me know if I can help with anything. Thanks!

src/components/metrics.tsx
zhuje: In the latest commit, removed the `react-csv` library components and rewrote the PR. I couldn't get the library to handle some data syncing...so I just decided to remove the library. Here's the snippet that formats the data items with the enclosingCharacters `""`.

```
  const getCsv = (array, delimiter = ',') =>
    array
      .map((row) =>
        row.map((rowItem) => (isNaN(rowItem) ? `"${rowItem}"` : rowItem)).join(delimiter),
      )
      .join('\n');
```
> `"${rowItem}"` -- wraps strings with the enclosingCharacters `""`

src/components/metrics.tsx
zhuje: Good advice, makes sense! I updated this in the latest commit so that the CSV is only generated `onClick` when the user selects 'Export as CSV' from the Kebab menu.

src/components/metrics.tsx
zhuje: Thanks for the suggested function, I tested it and it works! I updated this in the latest commit.

src/components/metrics.tsx
zhuje: Ah, yes, that makes sense. In the latest commit, I removed the icon and followed the mock-ups for Option 2. The 'Export as CSV' option is now a DropdownItem in the QueryKebab menu. 

 Also, I linked mock-up JIRAs ([OU-235](https://issues.redhat.com/browse/OU-235) and [OU-237](https://issues.redhat.com/browse/OU-237)) to the respective dev JIRAs ([OU-236](https://issues.redhat.com/browse/OU-236) and [OU-238](https://issues.redhat.com/browse/OU-238)).

src/components/metrics.tsx
zhuje: Ah, I see! Thank you for the correction. I updated this in the latest commit.

PR #110, opened 2024-03-27
Disable nginx module for rhel9

joepvd:


Tai-RedHat:
PR merged, will test with payload

PR #111, opened 2024-04-01
OCPBUGS-12857: Gracefully Handle NaN values in stacked graphs

PeterYurkovich:
This PR looks to address a crash for stacked charts when NaN values are returned from the Prometheus API.

There is a bug in stacked graphs for Victory charts for react. The bug occurs in their victory-stack package. An issue has been made and can be tracked [here](https://github.com/FormidableLabs/victory/issues/2844).

The root issue is a dataset which has dates in the x field and all null's in the y field. To get around this, for stacked charts we will be converting any NaN's returned to be 0 rather than null. In order to help dashboard creators and power users to debug any problematic queries, we are also adding a console.warn to let them know about this conversion, while still showing the graph otherwise.

### DIFF
--- a/src/components/query-browser.tsx
+++ b/src/components/query-browser.tsx
@@ -499,12 +499,13 @@ const formatSeriesValues = (
   values: PrometheusValue[],
   samples: number,
   span: number,
+  defaultEmptyValue: 0 | null,
 ): GraphDataPoint[] => {
   const newValues = _.map(values, (v) => {
     const y = Number(v[1]);
     return {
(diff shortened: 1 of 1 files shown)


jgbernalp:
LGTM, @juzhao feel free to add the lgtm label when regressions are checked

juzhao:
test along with along with https://github.com/openshift/console/pull/13712, stacked Graphs is handling NaN well, steps see from https://issues.redhat.com/browse/OCPBUGS-12857
/label qe-approved

PR #124, opened 2024-06-25
OCPBUGS-35156: Update express to v.4.19.2 

zhuje:
Update `yarn.lock` to upgrade`express` to version 4.19.2.


jgbernalp:
/hold

We don't use express, this might be a dev dependency

zhuje:
> /hold
> 
> We don't use express, this might be a dev dependency

Ah, you're right. I'll mark close the PR and close the JIRA issue to indicate it's a dev dependency.

PR #125, opened 2024-06-27
OU-251: Migrate Developer Perspective Alert List to Monitoring Plugin

PeterYurkovich:
This PR is apart of a pair of PR's (along with [this](https://github.com/openshift/console/pull/14014) one in the console codebase) which look to move the alerting list page from the console codebase over to the monitoring one. This PR needs to merge before the PR in the console codebase can merge. I would further prefer if both could be merged at the same time, by receiving approval on both before either merges.

There is a good amount of code that has changed and a good amount that has been split into other files. While reviewing the code, if you get lost then feel free to take a look at the different commits for this PR, as they are grouped by what they do. Once this PR is approved I will squash them all down, and then re-request approval.

This PR changes the UI of the developer perspective to be consistent with the admin perspective. The overall functionality stays the same, but there are some changes to the workflows required. For example to silence the alert you can't do so directly from the table anymore, instead you will click the ellipsis and then select the "Silence alert" option.

Screenshot 2024-06-27 at 1 36 14 PM

In addition, the tab control is done manually in the console block, with new additions being added at the end. So until the rest of the monitoring tabs are moved over the tabs will be out of their typical order as seen in the image above.

### DIFF
--- a/src/components/alerting.tsx
+++ b/src/components/alerting.tsx
@@ -1,17 +1,14 @@
 import {
-  Action,
   ActionServiceProvider,
   Alert,
   AlertSeverity,
   AlertStates,
-  BlueInfoCircleIcon,
   consoleFetchJSON,
   GreenCheckCircleIcon,
   ListPageFilter,
   PrometheusAlert,
   PrometheusEndpoint,
   PrometheusLabels,
-  RedExclamationCircleIcon,
   ResourceLink,
   ResourceStatus,
   RowFilter,
@@ -21,11 +18,9 @@ import {
   SilenceStates,
   TableColumn,
   Timestamp,
-  useActivePerspective,
   useListPageFilter,
   useResolvedExtensions,
   VirtualizedTable,
-  YellowExclamationTriangleIcon,
 } from '@openshift-console/dynamic-plugin-sdk';
 import {
   Alert as PFAlert,
@@ -51,13 +46,7 @@ import {
   ToolbarGroup,
   ToolbarItem,
 } from '@patternfly/react-core';
-import {
-  BanIcon,
-  BellIcon,
-  BellSlashIcon,
-  HourglassHalfIcon,
-  OutlinedBellIcon,
-} from '@patternfly/react-icons';
+import { BanIcon, HourglassHalfIcon } from '@patternfly/react-icons';
 import { sortable } from '@patternfly/react-table';
 import classNames from 'classnames';
 import { TFunction } from 'i18next';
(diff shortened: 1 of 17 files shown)


src/components/alerting/AlertsPage.tsx
```diff
@@ -0,0 +1,310 @@
...
+      type: 'alert-source',
+    },
+  ];
+
+  if (isDev) {
+    rowFilters = rowFilters.filter((filter) => filter.type !== 'alert-source');
+  }
+
+  const [staticData, filteredData, onFilterChange] = useListPageFilter(
+    data?.filter((alert: Alert) => (isDev ? alertSource(alert) === AlertSource.User : true)),
```
jgbernalp, line 122: This will prevent admins from using the dev console to see the namespace alerts. Although this behavior could be revised, at this stage, we need to maintain parity with the current dev console alerts.

jgbernalp (changes requested):

src/components/alerting.tsx
```diff
@@ -2408,10 +2013,20 @@ const PollerPages = () => {
...
-  }, [alertsSource, dispatch]);
+    return (): void => {
+      _.each(pollerTimeouts, clearTimeout);
+    };
+  }, [alertsSource, dispatch, perspective, rulesKey, alertsKey, namespace]);
+
+  if (isDev) {
+    return (
+      <Switch>
+        <Route path="/dev-monitoring/ns/:ns/(alerts)" exact component={AlertsPage} />
```
jgbernalp, line 2026: I think we don't need the capture group as is a single option
```suggestion
        <Route path="/dev-monitoring/ns/:ns/alerts" exact component={AlertsPage} />
```

jgbernalp (changes requested):

src/components/alerting.tsx
PeterYurkovich: Remove the capture group in the most recent commit

src/components/alerting/AlertsPage.tsx
PeterYurkovich: Removed the filter here. While testing I got mixed up by the console returning non-namespaced platform alerts while in "no authentication" mode for the rules tenancy api with a namespace, but after some more testing that is how it worked before this change as well. It should now have parity with the dev perspective alerts.

PeterYurkovich:
I've also gone ahead and squashed the commits down to a single one so that it is ready to be merged

jgbernalp (approved):

PR #126, opened 2024-06-27
OU-248: Split code from Alert Details Page out into other files

PeterYurkovich:
The original story for this was to move the Alerts Details page to the monitoring-plugin, however that is already the case. In keeping with the rest of this set of stories, this PR instead looks to split the code from the `alerting.tsx` file out into the individual pages as well as util files which contain shared code across the different pages.

This PR is currently branched off of #125 and cannot be merged until that is merged. Once merged, this will need to be rebased onto main. For review, only the last 2 commits are exclusive to this branch.

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -78,5 +78,16 @@
       "section": "observe",
       "insertAfter": "dashboards"
     }
+  },
+  {
+    "type": "console.tab",
+    "properties": {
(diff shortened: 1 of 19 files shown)


PeterYurkovich:
I will remove the hold on this PR after #125 has merged and this branch has been rebased to target the main branch instead.

PR #127, opened 2024-06-28
OU-238: Export CSV

krishagarwal278:


src/components/dashboards/index.tsx
```diff
@@ -613,6 +617,99 @@ const Card: React.FC<CardProps> = React.memo(({ panel }) => {
...
+            continue;
+          }
+          //const series = item[0][name].toString()
+          result[dateTime].push({ [name]: value });
+        }
+      }
+    }
+    const firstDateTime = Object.keys(result)[0];
+
+    const uniqueVerbs = new Set();
```
jgbernalp, line 679: This only applies for one chart example, would it be better to have?
```suggestion
    const uniqueSeries = new Set();
```

src/components/dashboards/index.tsx
```diff
@@ -613,6 +617,99 @@ const Card: React.FC<CardProps> = React.memo(({ panel }) => {
...
+  const [csvData, setCsvData] = React.useState([]);
+
+  const csvExportHandler = () => {
+    let csvString = '';
+    const result = {};
+    const nameArray = []; //
+
+    for (const [i, row] of csvData.entries()) {
+      for (const item of row) {
+        for (const entry of item[1]) {
```
jgbernalp, line 661: It seems we are just iterating over the first element all the time, should it be?
```suggestion
        for (const entry of row[0][1]) {
```

src/components/dashboards/index.tsx
```diff
@@ -613,6 +617,99 @@ const Card: React.FC<CardProps> = React.memo(({ panel }) => {
...
+    },
+    [panel],
+  );
+
+  const [csvData, setCsvData] = React.useState([]);
+
+  const csvExportHandler = () => {
+    let csvString = '';
+    const result = {};
+    const nameArray = []; //
```
jgbernalp, line 657: Are these the column names?

src/components/dashboards/index.tsx
```diff
@@ -613,6 +617,99 @@ const Card: React.FC<CardProps> = React.memo(({ panel }) => {
...
+          //const series = item[0][name].toString()
+          result[dateTime].push({ [name]: value });
+        }
+      }
+    }
+    const firstDateTime = Object.keys(result)[0];
+
+    const uniqueVerbs = new Set();
+
+    result[firstDateTime].forEach((item) => {
```
jgbernalp, line 681: Are we assuming that the first row has all the series, could it be that the first row does not have values in all the series?

src/components/dashboards/index.tsx
```diff
@@ -613,6 +617,99 @@ const Card: React.FC<CardProps> = React.memo(({ panel }) => {
...
+      const verb = Object.keys(item)[0];
+      uniqueVerbs.add(verb);
+    });
+
+    const uniqueVerbsArray = Array.from(uniqueVerbs);
```
jgbernalp, line 686: When changing a set to an array the order might not be preserved, leading to put values in the wrong series

src/components/dashboards/index.tsx
```diff
@@ -734,17 +815,21 @@ const Card: React.FC<CardProps> = React.memo(({ panel }) => {
...
-                      pollInterval={pollInterval}
-                      queries={queries}
-                      showLegend={panel.legend?.show}
-                      units={panel.yaxes?.[0]?.format}
-                      onZoomHandle={handleZoom}
-                      namespace={namespace}
-                      customDataSource={customDataSource}
-                    />
+                    <>
+                      {queries}
```
jgbernalp, line 819: Are the queries needed here?

krishagarwal278: removed queries as this was only added to see the the data that's being called on the cards in the dashboard

src/components/dashboards/index.tsx
krishagarwal278: yes, first row has all the series, in all the graphs I've witnessed so far. It has the same key-value pair stored in some sort of an array

src/components/dashboards/index.tsx
krishagarwal278: these are the names of the different series

src/components/dashboards/index.tsx
krishagarwal278: I'm getting further errors when I change it to the suggested change, with labels[key] as undefined. However, it's working properly with two loops

src/components/dashboards/index.tsx
jgbernalp: then the name could be `seriesNames`

src/components/dashboards/index.tsx
jgbernalp: I'm not sure if this would be a general rule, we need to account for this edge case

Dockerfile
```diff
@@ -27,4 +27,4 @@ USER 1001
 
 COPY --from=builder /usr/src/app/dist /usr/share/nginx/html
 
-ENTRYPOINT ["nginx", "-g", "daemon off;"]
+ENTRYPOINT ["nginx", "-g", "daemon off;"]
```
jgbernalp, line 30:
```suggestion
ENTRYPOINT ["nginx", "-g", "daemon off;"]

```

src/components/dashboards/index.tsx
```diff
@@ -57,6 +60,7 @@ import IntervalDropdown from '../poll-interval-dropdown';
 import { RootState } from '../types';
 import BarChart from './bar-chart';
 import Graph from './graph';
+//import graphData from './query-browser.tsx'
```
jgbernalp, line 63: Please remove commented or unused code as this code must be production ready

src/components/dashboards/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
```
jgbernalp, line 1: We should not disable lint checkers
```suggestion
```

src/components/dashboards/index.tsx
```diff
@@ -613,6 +618,82 @@ const Card: React.FC<CardProps> = React.memo(({ panel }) => {
...
+        /{{\s*([a-zA-Z_][a-zA-Z0-9_]*)\s*}}/g,
+        (match, key) => labels[key] ?? '',
+      );
+    },
+    [panel],
+  );
+  const [csvData, setCsvData] = React.useState([]);
+
+
+const csvExportHandler = () => {
```
jgbernalp, line 640: indenting and formatting in this function is missing, please re enable prettier to make the code formatting consistent

yarn.lock
jgbernalp, line 1: did you use `npm` instead of `yarn`? This project uses `yarn` and changes to this file might be incorrect

src/components/dashboards/index.tsx
```diff
@@ -711,6 +775,7 @@ const Card: React.FC<CardProps> = React.memo(({ panel }) => {
             {!isLoading && (
               <QueryBrowserLink queries={queries} customDataSourceName={customDataSourceName} />
             )}
+            {panel.type === 'graph' && <KebabDropdown dropdownItems={dropdownItems} />}
```
jgbernalp, line 778: This should be hidden when we have no data

jgbernalp (changes requested):

src/components/dashboards/index.tsx
```diff
@@ -698,6 +756,16 @@ const Card: React.FC<CardProps> = React.memo(({ panel, perspective }) => {
...
   const panelClassModifier = getPanelClassModifier(panel);
 
+  const isThereCsvData = () => {
+    const result = false;
+    if (csvData.length > 0) {
+      if (csvData[0].length > 0) {
+        return true;
+      }
+    }
+    return result;
```
jgbernalp, line 766:
```suggestion
    if (csvData.length > 0) {
      if (csvData[0].length > 0) {
        return true;
      }
    }
    return false;
```

src/components/dashboards/types.ts
```diff
@@ -21,6 +21,7 @@ type YAxis = {
 };
 
 export type Panel = {
+  data: boolean;
```
jgbernalp, line 24: Do we need this data field, and also should it be a boolean?

start-console.sh
```diff
@@ -27,12 +27,12 @@ if [ -x "$(command -v podman)" ]; then
     if [ "$(uname -s)" = "Linux" ]; then
         # Use host networking on Linux since host.containers.internal is unreachable in some environments.
         BRIDGE_PLUGINS="${npm_package_consolePlugin_name}=http://localhost:9001"
-        podman run --pull always --rm --network=host --env-file <(set | grep BRIDGE) $CONSOLE_IMAGE
+        podman run --pull always --arch=amd64 --rm --network=host --env-file <(set | grep BRIDGE) $CONSOLE_IMAGE
     else
         BRIDGE_PLUGINS="${npm_package_consolePlugin_name}=http://host.containers.internal:9001"
-        podman run --pull always --rm -p "$CONSOLE_PORT":9000 --env-file <(set | grep BRIDGE) $CONSOLE_IMAGE
+        podman run --pull always --arch=amd64 --rm -p "$CONSOLE_PORT":9000 --env-file <(set | grep BRIDGE) $CONSOLE_IMAGE
```
jgbernalp, line 33: This should probably be a variable at this does not apply to every developer environment, something like: https://github.com/openshift/troubleshooting-panel-console-plugin/blob/main/scripts/start-console.sh#L8

jgbernalp (changes requested):

start-console.sh
```diff
@@ -21,18 +22,23 @@ BRIDGE_USER_SETTINGS_LOCATION="localstorage"
...
         BRIDGE_PLUGINS="${npm_package_consolePlugin_name}=http://localhost:9001"
-        podman run --pull always --rm --network=host --env-file <(set | grep BRIDGE) $CONSOLE_IMAGE
+        podman run --pull always --platform $CONSOLE_IMAGE_PLATFORM --rm --network=host --env-file <(set | grep BRIDGE) $CONSOLE_IMAGE
     else
         BRIDGE_PLUGINS="${npm_package_consolePlugin_name}=http://host.containers.internal:9001"
-        podman run --pull always --rm -p "$CONSOLE_PORT":9000 --env-file <(set | grep BRIDGE) $CONSOLE_IMAGE
+        podman run --pull always --platform $CONSOLE_IMAGE_PLATFORM \
+            --rm -p "$CONSOLE_PORT":9000 \
+            --env-file <(set | grep BRIDGE) \
+            --env BRIDGE_PLUGIN_PROXY='{"services": [{"consoleAPIPath": "/api/proxy/plugin/troubleshooting-panel-console-plugin/korrel8r/", "endpoint":"https://host.containers.internal:9005","authorize":true}]}' \
```
PeterYurkovich, line 38: The monitoring plugin doesn't need a proxy to korrel8r

jgbernalp (approved):

PR #128, opened 2024-07-01
OU-436: Add Golang backend

PeterYurkovich:
This PR looks to add a lightweight Golang backend for the monitoring-plugin. It adds some standard features used in the rest of the Observability UI team's repos such as a script folder including a `build-image.sh` script and a makefile with simple commands.

This PR is currently lacking setup for the `Dockerfile.ci` and the `Dockerfile.art`, however other Dockerfiles are updated and can be used for testing. In addition these changes will not work with downstream items using the monitoring-plugin such as the [cluster-monitoring-operator](https://github.com/openshift/cluster-monitoring-operator) due to the internal changes of the built image, which will be addressed in [OU-450](https://issues.redhat.com/browse/OU-450).

The backend adds a timeout parameter that is standard across the Observability UI team's repos, but may not be needed here. If desired it can be removed from this PR.


PeterYurkovich:
Putting a hold onto this as I don't think it can be merged until [OU-450](https://issues.redhat.com/browse/OU-450) is completed, but it's ready for review within the confines of this project

PeterYurkovich:
When testing, it would be best to delete your `node_modules` or `dist` folder if you have one first since the `.dockerignore` and `.gitignore` have been updated to ignore the ones in the `web` folder now and not the old ones so your git and docker might struggle with the excessive "new" data of those folders in the old locations

.husky/pre-commit
jgbernalp, line 1: this file should be probably inside `web`

web/package.json
```diff
@@ -15,10 +15,9 @@
     "lint": "eslint ./src",
     "lint-fix": "eslint ./src --fix",
     "pre-commit:check": "yarn lint",
-    "prepare": "husky install",
```
jgbernalp, line 18: we should keep this when the .husky folder is inside `web`

.husky/pre-commit
PeterYurkovich: Moved the husky files into the `web`

web/package.json
PeterYurkovich: fixed the prepare script so that it now works within a subdirectory

nginx.conf
PeterYurkovich, line 1: Is this file needed in this PR? It doesn't look like it's used anywhere so unless its automatically pulled in by nginx or is going to be use by downstream projects we can remove it

Dockerfile.art
```diff
@@ -36,4 +45,10 @@ USER 1001
 
 COPY --from=builder /usr/src/app/dist /usr/share/nginx/html
```
PeterYurkovich, line 46: This needs to be changed to use `web-builder` instead as `builder` no longer exists

I agree with duplicating the dist and putting it in both the `/usr/share/nginx/html` and at the `/opt/app-root/web`. It should make it a lot easier to transition to just the go binary

PeterYurkovich (commented):
I can't request changes on my own pull request, but the `Dockerfile.art` needs one change and I have a question on a conf file being added

nginx.conf
jgbernalp: no is not, part of my local tests. Is removed now.

Dockerfile.art
jgbernalp: Thnx for the suggestion, indeed is the `web-builder` now, is fixed now

PR #129, opened 2024-07-03
OU-439: Add the predefined queries list to the admin metrics 

zhuje:
On the metrics page, add a predefined queries dropdown.

image

### DIFF
--- a/src/components/metrics.tsx
+++ b/src/components/metrics.tsx
@@ -20,6 +20,11 @@ import {
   EmptyStateVariant,
   Switch,
   Title,
+  Select,
+  SelectVariant,
+  SelectOption,
+  Grid,
+  GridItem,
 } from '@patternfly/react-core';
 import {
   AngleDownIcon,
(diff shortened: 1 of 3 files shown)


PeterYurkovich (changes requested):
In our sprint planning we had discussed placing the dropdown to the left of the "Add query" and "Run queries" buttons under the graph. If we decided otherwise thats fine and you can ignore this, but as of our sprint planning discussion the dropdown should be moved down under the graph.

src/components/metrics.tsx
```diff
@@ -84,6 +89,123 @@ import {
 // Stores information about the currently focused query input
 let focusedQuery;
 
+type PredefinedQueryType = {
+  name: string;
+  query: string;
+};
+
+export const PreDefinedQueriesDropdown = () => {
+  const [isOpen, setIsOpen] = React.useState(false);
```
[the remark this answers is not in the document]
zhuje: I'm going to leave it as a React.useState(). After attempting to switch to the useBoolean() hook the <Select> component couldn't open anymore.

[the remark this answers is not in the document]
PeterYurkovich: After debugging a bit, the `<Select>` from Patternfly 4 fires the `onToggle` function twice when trying to open, leading to the `useBoolean` hook to loop back to the original `closed` state, so the `useState` is fine here. @zhuje added a comment to note this in the codebase in case further actions are needed to be taken on opening the dropdown

src/components/metrics.tsx
```diff
@@ -84,6 +89,126 @@ import {
...
+    {
+      name: 'Rate of transmitted packets dropped',
+      query: `sum(irate(container_network_transmit_packets_dropped_total[2h])) by (pod)`,
+    },
+  ];
+
+  // Note this fires twice when <Select> is clicked.
+  const onToggle = () => {
+    setIsOpen(!isOpen);
+  };
```
jgbernalp, line 148:
```suggestion
  const onToggle = (isOpen:boolean) => {
    setIsOpen(isOpen);
  };
```

jgbernalp (changes requested):

src/components/metrics.tsx
```diff
@@ -84,6 +89,126 @@ import {
...
+  );
+
+  const insertPredefinedQuery = (query: string) => {
+    const queriesListDetails = queriesList.toJS();
+    const isInitialQueryEmpty =
+      queriesList.size === 1 &&
+      (queriesListDetails[0]?.text === '' ||
+        queriesListDetails[0]?.text === null ||
+        queriesListDetails[0]?.text === undefined);
+    const index = isInitialQueryEmpty ? 0 : queriesList.size;
```
jgbernalp, line 163: we might want to insert the query only if its different from the current first

jgbernalp (changes requested):

src/components/metrics.tsx
zhuje: Updated in the latest commit.

zhuje:
/cc @wzheng1

@wzheng1 when you have time could you take a look at this?

PR #131, opened 2024-07-09
OU-248: Add Alerts Detail Page

PeterYurkovich:
[ODC-7640](https://issues.redhat.com/browse/ODC-7640)
[OU-248](https://issues.redhat.com/browse/OU-248)

This PR is the counterpart of one in the console ([PR](https://github.com/openshift/console/pull/14072)) which looks to transfer the dev perspective alerts detail page out of the console codebase and into the monitoring-plugin. This PR must merge before the one in the console has.

image

The only change to the dev perspective alerts detail page is that it will no longer have a namespace selector. Since the selected alert is already namespaced, it didn't make sense to swap namespaces while focused on a specific alert.


PeterYurkovich:
Upon further review, due to the `fetch-alerts.tsx` file combining a number of different API calls, I don't think it would be best to refactor the `useRulesAlertsPoller` to use the `useURLPoll`. I believe that we should keep them as separate polling implementations for now

PeterYurkovich:
Unlike the alerts list stories when this PR is merged but the one in the console isn't merged, the alerts detail page will be unavailable. I'm putting a hold on this story so that QE could take a look at both parts together before this merges

juzhao:
tested with PRs
`launch 4.17.0-0.nightly-2024-07-17-183402,openshift/monitoring-plugin#131,openshift/console#14072 aws`
there is one bug: https://issues.redhat.com/browse/OCPBUGS-37232

PR #134, opened 2024-07-15
OCPBUGS-36970: Use the non-tenancy URL in the Admin perspective

PeterYurkovich:
Since the namespace variable is what determines whether the tenancy URL or the admin URL is used in the `getPrometheusURL` we should ensure that it only gets passed in when being called in the dev perspective.


juzhao:
@Tai-RedHat please help to test with PR to see if there are regression issues

Tai-RedHat:
test with cluster-bot, alerting pages back. Also checked `Metrics/Dashboards/Targets` pages, no other regression issues.
/label qe-approved

PR #137, opened 2024-07-25
OCPBUGS-37291: check undefined function before calling

jgbernalp:


andreasgerstmayr (approved):

src/components/query-browser.tsx
```diff
@@ -847,7 +847,7 @@ const QueryBrowser_: React.FC<QueryBrowserProps> = ({
             },
           );
           setGraphData(newGraphData);
-          onDataChange(newGraphData);
+          onDataChange?.(newGraphData);
```
andreasgerstmayr, line 850: I'm curious why this didn't throw an TypeScript error? It's correctly marked as optional in `QueryBrowserProps`.

PR #140, opened 2024-08-05
added fetch for alerts and data processer

Fewwy:
I am adding fetch for alerts so I can process the data and prepare it for the graph
I added a data processer that converts timestamps into dates readable by human

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4,24 +4,24 @@
 
 "@aashutoshrathi/word-wrap@^1.2.3":
   version "1.2.6"
-  resolved "https://registry.npmjs.org/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz"
+  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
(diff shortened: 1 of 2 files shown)


src/components/console/utils/safe-fetch-hook.ts
```diff
@@ -9,5 +9,5 @@ export const useSafeFetch = () => {
   }, []);
 
   return (url) =>
-    consoleFetchJSON(url, 'get', { signal: controller.current.signal as AbortSignal });
+    consoleFetchJSON(url, 'get', { signal: controller.current?.signal as AbortSignal });
```
Fewwy, line 12: @zhuje @kyoto For some reason this is undefined so I had to add an optional chain here. Do you have any idea why it is happening?

src/components/Incidents/IncidentsPage.jsx
```diff
@@ -1,9 +1,107 @@
...
+      const promiseMap = _.map(timeRanges, (timeRange) =>
+        safeFetch(
+          getPrometheusURL(
+            {
+              endpoint: PrometheusEndpoint.QUERY_RANGE,
+              endTime: timeRange.endTime,
+              namespace,
+              query,
+              samples: Math.ceil(samples / timeRanges.length),
+              timeout: '60s',
```
jgbernalp, line 66: should we leave this empty so the timeout defaults to the one configured in prometheus?

yarn.lock
jgbernalp, line 1: Did you use npm instead of yarn? We will migrate out of yarn, but the current code still uses yarn

PR #148, opened 2024-08-12
refactored fetch connected it to data processing

Fewwy:
Removed promises fetch and changed it to an easier-to-read async await.
Added data processing that waits for data to be fetched then changes data into format of HH/DD/MM/YY

### DIFF
--- a/src/components/Incidents/AlertsChart/AlertsChart.jsx
+++ b/src/components/Incidents/AlertsChart/AlertsChart.jsx
@@ -10,6 +10,9 @@ import {
 } from '@patternfly/react-charts';
 import { VictoryZoomContainer } from 'victory-zoom-container';
 import { Card, CardTitle } from '@patternfly/react-core';
+import global_danger_color_100 from '@patternfly/react-tokens/dist/esm/global_danger_color_100';
+import global_info_color_100 from '@patternfly/react-tokens/dist/esm/global_info_color_100';
+import global_warning_color_100 from '@patternfly/react-tokens/dist/esm/global_warning_color_100';
 
 const days = [
   'Jul 25',
@@ -30,11 +33,7 @@ const days = [
 ];
 const AlertsChart = () => {
   return (
-    <Card
-      style={{
-        border: '6px solid #F3F3F3',
-      }}
-    >
+    <Card className="alerts-chart-card">
       <CardTitle>Alerts Timeline</CardTitle>
       <div
         style={{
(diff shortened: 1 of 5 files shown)


src/components/Incidents/AlertsChart/AlertsChart.jsx
```diff
@@ -1,140 +1,167 @@
...
+  'Aug 6',
+  'Aug 7',
+  'Today',
+];
 const AlertsChart = () => {
   return (
-    <Card>
+    <Card
+      style={{
+        border: '6px solid #F3F3F3',
```
jgbernalp, line 35: This should probably come from patternfly CSS vars as this will need to change if user is in dark mode

src/components/Incidents/AlertsChart/AlertsChart.jsx
```diff
@@ -1,140 +1,167 @@
...
+  'Aug 1',
+  'Aug 2',
+  'Aug 3',
+  'Aug 4',
+  'Aug 5',
+  'Aug 6',
+  'Aug 7',
+  'Today',
+];
 const AlertsChart = () => {
```
jgbernalp, line 31: It seems that this component is the same as Incidents chart. Would you say they can be reused or they will diverge enough in order to have 2 separate components?

src/components/Incidents/IncidentsPage.jsx
```diff
@@ -49,57 +54,33 @@ const IncidentsPage = ({
...
-        safeFetch(
+  React.useEffect(() => {
+    (async () => {
+      try {
+        const response = await safeFetch(
           getPrometheusURL(
             {
               endpoint: PrometheusEndpoint.QUERY_RANGE,
-              endTime: timeRange.endTime,
+              endTime: 1722950372519,
```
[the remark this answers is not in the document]
Fewwy: Right now this property is fixed to make sure that it works as expected.
Yes, My plan is to pass time range via filters in the toolbar later. I added moment js functions to the utils to "find" the date ranges correctly. I will test them soon. 
image

src/components/Incidents/AlertsChart/AlertsChart.jsx
Fewwy: There is no mocks for darkmode. so what should I do with it?

jgbernalp: Patternfly components support dark mode so we should use the CSS variables which already adjust to it instead of using hardcoded colors. In this specific case we might not need to change the color, just the border width

PR #154, opened 2024-08-13
NO-JIRA: Use DNF and go based image in Dockerfile

PeterYurkovich:
The okd-scos image in 4.17 being used doesn't have access to microdnf. This PR adds a slightly heavier image, but one that has access to dnf to install the needed dependencies.

Testing for this PR on the 4.17 branch is being done here in #153

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,9 +30,9 @@ COPY pkg/ pkg/
 
 RUN make build-backend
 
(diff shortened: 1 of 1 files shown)


PeterYurkovich:
Creating this PR ahead of of the 
#153 testing being completed so that the CI can be completed for this one once that the testing is done

PR #157, opened 2024-08-14
NO-JIRA: add FIPS compliant flags and env vars

jgbernalp:


juzhao:
@Tai-RedHat please help to test with FIPS enabled/disabled cluster after the code merged to 4.18 nightly payload

PR #168, opened 2024-08-27
Improved fetch and data preparation for charts

Fewwy:
1. Removed a bunch of unnecessary code, now it is cleaner and easier to understand
2. Updated fetch function: now it collects the data depending on the number of days requested and sets it into state
3. Added functions to combine data, dedupe it and prepare it to be consumed by chart

### DIFF
--- a/src/components/Incidents/AlertsChart/AlertsChart.jsx
+++ b/src/components/Incidents/AlertsChart/AlertsChart.jsx
@@ -56,7 +56,7 @@ const AlertsChart = ({ alertsData }) => {
               />
             }
             //TODO: domain ranges should be adjusted based on the amount of rows and day range
-            domain={{ x: [0, 17], y: [0, 15] }}
+            domain={{ x: [0, 2], y: [0, 15] }}
             domainPadding={{ x: [30, 25] }}
             legendData={[
               { name: 'Critical', symbol: { fill: global_danger_color_100.var } },
@@ -64,6 +64,7 @@ const AlertsChart = ({ alertsData }) => {
               { name: 'Warning', symbol: { fill: global_warning_color_100.var } },
             ]}
             legendPosition="bottom-left"
+            //this should be always less than the container height
             height={300}
             padding={{
               bottom: 75, // Adjusted to accommodate legend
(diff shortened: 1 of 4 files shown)


src/components/Incidents/AlertsChart/AlertsChart.jsx
```diff
@@ -7,167 +7,187 @@ import {
...
-  'Aug 3',
-  'Aug 4',
-  'Aug 5',
-  'Aug 6',
-  'Aug 7',
-  'Today',
-];
-const AlertsChart = () => {
+const AlertsChart = ({ alertsData }) => {
+  const [isLoading, setIsLoading] = React.useState(true);
```
PeterYurkovich, line 21: Rather than trying to determine what is happening in a higher level component based on the props being passed down, imo we should instead just keep track of the loading state in the `IncidentsPage` component and then pass it down to this component.

src/components/Incidents/IncidentsPage.jsx
```diff
@@ -10,67 +10,62 @@ import { useDispatch } from 'react-redux';
...
-  }, [span, fixedXDomain]);
-  const endTime = xDomain?.[1];
-
-  React.useEffect(() => {
-    if (timespan) {
-      setSpan(timespan);
-      setSamples(defaultSamples || getMaxSamplesForSpan(timespan));
-    }
-  }, [defaultSamples, timespan]);
+  const [xDomain, setXDomain] = React.useState(twoWeeksDateRanges);
```
PeterYurkovich, line 44: Is there a reason other than using test data that is consistent that the old "find 2 weeks of dates" was removed? If there is then don't worry about this comment, but if it's just going to have to come back in a future PR then we should just leave it even just commented out or something.

src/components/Incidents/IncidentsPage.jsx
```diff
@@ -10,67 +10,62 @@ import { useDispatch } from 'react-redux';
...
+            );
+            return response.data.result;
+          }),
         );
-        setData(response.data.result);
+        const aggregatedData = results.reduce((acc, result) => acc.concat(result), []);
+        setData(aggregatedData);
       } catch (err) {
         // eslint-disable-next-line no-console
         console.log(err);
```
PeterYurkovich, line 71: When working with promises, you can use `.then()`, `.catch()`, and `.finally()` on the promise itself to help handle the results of it rather than keeping track of scopes of try catch blocks and awaits. It is mostly a code formatting preference, but that is how we tend to have our code in our plugins

```suggestion
        Promise.all(
          twoWeeksDateRanges.map(async (range) => {
            const response = await safeFetch(
              getPrometheusURL(
                {
                  endpoint: PrometheusEndpoint.QUERY_RANGE,
                  endTime: range.endTime,
                  namespace,
                  query: 'ALERTS',
                  samples: 23,
                  timespan: range.duration - 1,
                },
                customDataSource?.basePath,
              ),
            );
            return response.data.result;
          }),
        ).then((results) => {
          const aggregatedData = results.reduce((acc, result) => acc.concat(result), []);
          setData(aggregatedData);
        })
        .catch((err) => {
          // eslint-disable-next-line no-console
          console.log(err);
        })
```

src/components/Incidents/IncidentsPage.jsx
```diff
@@ -81,11 +76,10 @@ const IncidentsPage = ({
   React.useEffect(() => {
     setProcessedData(processIncidentsTimestamps(data));
   }, [data]);
```
PeterYurkovich, line 78: Rather than adding an extra use effect we can just set the processed data in the above useEffect

PeterYurkovich (changes requested):

src/components/Incidents/IncidentsPage.jsx
Fewwy: This is a temporary thing and will be changed in the next PR

PR #171, opened 2024-08-28
NO-JIRA: Add `make install` command to install frontend and backend dependencies

zhuje:


PeterYurkovich:
I think you should go ahead and tag this PR with https://issues.redhat.com/browse/OU-487 so all of the PR's on the developer tooling you might do show up in jira.

zhuje:
I don't think this should be tagged with  https://issues.redhat.com/browse/OU-487. This issue is not really centralizing tooling under the observability-ui repo. 

I'm going to leave it as NO-JIRA and unhold.

PR #179, opened 2024-09-05
OU-450: Use plugin-backend as default ENTRYPOINT

PeterYurkovich:
This PR looks to remove the backwards compatibility of the `monitoring-plugin` images and instead use the go backend


PeterYurkovich:
Doing test builds with ART

PR #180, opened 2024-09-06
First Pass on listing and connecting to alert managers

PeterYurkovich:
This PR starts the work on the ACM branch for the work needed on on both the frontend and the backend.

Rather than fighting roles and role bindings manually I have taken a bit of extra time to skip ahead to [OU-407](https://issues.redhat.com/browse/OU-407) and start on a branch [here](https://github.com/PeterYurkovich/observability-operator/tree/acm) which can be used for deployment.

The work and the work in ObO is in an incomplete state, and may not work. Reviews are welcome and once work is complete to a working order a review will be requested manually.


PeterYurkovich:
This PR looks at the ACM branch, not main. I would appreciate feedback on the alertmanager and thanos-querier paramaters being passed in. I considered a couple of options:
- placing them in a config file which is then read (which may be overkill)
- combining the name and namespace into a single variable (which is what is done here, but requires very specific formatting)
- leaving them in 4 separate variables (which creates a lot of variables)
- passing in the URLs which will be proxied too (which offloads business logic upstream somewhere to either COO or one of the ACM controllers)

None of them felt like particularly good solutions, so I'd be happy to discuss what would be best

charts/openshift-console-plugin/templates/clusterrole.yaml
```diff
@@ -0,0 +1,41 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: monitoring-plugin-cr
+rules:
+  - verbs:
```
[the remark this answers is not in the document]
PeterYurkovich: This was a remnant of when we were listing them. This is indeed no longer needed. I've removed it in the most recent commit.

charts/openshift-console-plugin/templates/deployment.yaml
```diff
@@ -31,19 +32,20 @@ spec:
...
-              mountPath: /etc/nginx/nginx.conf
-              subPath: nginx.conf
+          args:
+            - '-config-path=/opt/app-root/web/dist'
+            - '-static-path=/opt/app-root/web/dist'
+            - '-cert=/var/cert/tls.crt'
+            - '-key=/var/cert/tls.key'
+            - '-log-level=info'
+            - '-features=acm'
+            - '-alertmanager-location=main/openshift-monitoring'
```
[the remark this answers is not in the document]
PeterYurkovich: I have updated the entries to have separate name and namespace parameters. I also moved the proxies to separate ports, however I don't believe that we need to set up a custom CA for each one. So long as the traffic is routed through a single service then there shouldn't be any difference I don't believe. CMO does something similar for the thanos-querier (see [here](https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/thanos-querier/service.yaml).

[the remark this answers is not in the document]
PeterYurkovich: With the extra number of ports I didn't think it made too much sense to need to pass in the port numbers for all 3 ports as well. I left the main service as configurable, but made the proxy ports static.

pkg/server.go
```diff
@@ -118,14 +188,44 @@ func setupRoutes(cfg *Config) (*mux.Router, *PluginConfig) {
 	router := mux.NewRouter()
 
 	router.PathPrefix("/health").HandlerFunc(healthHandler())
+
+	// TODO: needs to check for acm feature and adjust the plugin-manifest to be something appropriate
 	router.Path("/plugin-manifest.json").Handler(manifestHandler(cfg))
```
jgbernalp, line 193: The manifest handler patches the manifest based on the features to add console extensions

cmd/plugin-backend.go
```diff
@@ -11,15 +11,19 @@ import (
...
-	log             = logrus.WithField("module", "main")
+	portArg                   = flag.Int("port", 0, "server port to listen on (default: 9443)\n9444 and 9445 reserved for other use")
+	certArg                   = flag.String("cert", "", "cert file path to enable TLS (disabled by default)")
+	keyArg                    = flag.String("key", "", "private key file path to enable TLS (disabled by default)")
+	featuresArg               = flag.String("features", "", "enabled features, comma separated")
+	staticPathArg             = flag.String("static-path", "", "static files path to serve frontend (default: './web/dist')")
+	configPathArg             = flag.String("config-path", "", "config files path (default: './config')")
+	pluginConfigArg           = flag.String("plugin-config-path", "", "plugin yaml configuration")
+	logLevelArg               = flag.String("log-level", "error", "verbosity of logs\noptions: ['panic', 'fatal', 'error', 'warn', 'info', 'debug', 'trace']\n'trace' level will log all incoming requests\n(default 'error')")
+	alertmanagerNameArg       = flag.String("alertmanager-name", "", "alertmanager name for acm mode")
```
jgbernalp, line 22: should we allow to provide the full URL rather than namespace and name?

```suggestion
	alertmanager       = flag.String("alertmanager", "", "alertmanager URL")
```

```
-alertmanager=https://alertmanager-main.openshift-monitoring.svc:9094
```

WDYT?

pkg/server.go
PeterYurkovich: The use case here will need to be a bit different, as ACM is going to need to trim down and change the features rather than add to them. This functionality falls into [OU-507](https://issues.redhat.com/plugins/servlet/mobile#issue/OU-507) which is the UI part, which I will be starting today. This PR is looking at a separate feature branch, so I don't see this as a blocker for this PR.

cmd/plugin-backend.go
PeterYurkovich: Most recent commit changes to use URL parameters. Also updates the feature flag to 'acm-alerting' rather than 'acm'.

PeterYurkovich:
Updates to the COO [PR](https://github.com/rhobs/observability-operator/pull/575) have also been completed to bring it in sync with changes to feature and parameter names.

PR #186, opened 2024-09-16
OCPBUGS-41602: Reduce default logging

PeterYurkovich:
Due to the addition of a health check on the monitoring plugin deployment, the `/health` endpoint is getting hit quite often, leading to an excessive amount of logs being generated. This is also true for other endpoints as well, such as the `/plugin-manifest.json` which is polled around every 15 seconds per user logged into the console.

This PR looks to combat these excessive logs by not logging individual requests when an endpoint is hit, unless a new `log-verbosity` parameter is set to 1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,33 +44,33 @@ Install the chart into a new namespace or an existing namespace as specified by 
 helm upgrade -i monitoring-plugin charts/openshift-console-plugin -n my-plugin-namespace --create-namespace --set plugin.image=my-plugin-image-location
 ```
 
-## Local Development 
+## Local Development
 
-### Dependencies 
+### Dependencies
 1. [yarn](https://yarnpkg.com/en/docs/install)
-2. [oc](https://mirror.openshift.com/pub/openshift-v4/clients/oc/4.4/) 
-3. [podman 3.2.0+](https://podman.io) or [Docker](https://www.docker.com) 
+2. [oc](https://mirror.openshift.com/pub/openshift-v4/clients/oc/4.4/)
+3. [podman 3.2.0+](https://podman.io) or [Docker](https://www.docker.com)
 4. An OpenShift cluster
 
-### Running Locally 
+### Running Locally
 ```
-# Login to an OpenShift cluster 
+# Login to an OpenShift cluster
 $ oc login <clusterAddress> -u <username> -p <password>
 
-# Start podman (or Docker) 
-$ podman machine init 
-$ podman machine start 
+# Start podman (or Docker)
+$ podman machine init
+$ podman machine start
 
-# Install dependencies 
-$ make install 
+# Install dependencies
+$ make install
 
-# Run the application 
+# Run the application
 $ make start-frontend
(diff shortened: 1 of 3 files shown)


cmd/plugin-backend.go
```diff
@@ -18,6 +18,7 @@ var (
 	staticPathArg   = flag.String("static-path", "", "static files path to serve frontend (default: './web/dist')")
 	configPathArg   = flag.String("config-path", "", "config files path (default: './config')")
 	pluginConfigArg = flag.String("plugin-config-path", "", "plugin yaml configuration")
+	logVerbosityArg = flag.Int("log-verbosity", 0, "verbosity of logs (0 for low, 1 for high)")
```
jgbernalp, line 21: it might be better to use a standard `log.level` flag with string values. These values can be set up directly in logrus, using something like:
```
level, err := logrus.ParseLevel(logLevel)
	if err != nil {
		logrus.WithError(err).Fatal("unable to set the log level")
	}
	logrus.SetLevel(level)
	```

Comment (PeterYurkovich) [reply cmd/plugin-backend.go]: I swapped to using the log-level strings like you had suggested. Since the request level logging is being done through a separate package, just changing the logrus level didn't prevent it, so that part of the code has been kept the same. I changed it instead to look for the "trace" level since that seemed the most apt.

File: pkg/server.go [hunk 2]
```diff
@@ -57,19 +56,32 @@ func Start(cfg *Config) {
...
 	httpServer := &http.Server{
-		Handler:      loggedRouter,
+		Handler:      router,
 		Addr:         fmt.Sprintf(":%d", cfg.Port),
 		TLSConfig:    tlsConfig,
 		ReadTimeout:  timeout,
 		WriteTimeout: timeout,
 	}
 
+	if logrusLevel == logrus.TraceLevel {
```
jgbernalp, line 73: isn't enough to set the log level? I think route logs will be on the info level so by default they'll be skipped

PeterYurkovich: The handlers.LoggingHandler comes from the gorilla/handlers package, which doesn't respect or use the logrus internally. Setting the log level for logrus has no impact on the route logs

jgbernalp: Ok got it. Thnx

README.md
```diff
@@ -44,33 +44,33 @@ Install the chart into a new namespace or an existing namespace as specified by
...
+# In a separate terminal
 $ make start-console
 ```
 The application will be running at [localhost:9000](http://localhost:9000/).
 
-### Local Development Troubleshooting 
+### Local Development Troubleshooting
 1. Disable cache. Select 'disable cache' in your browser's DevTools > Network > 'disable cache'. Or use private/incognito mode in your browser.
-
+2. Enable higher log verbosity by setting `-log-verbosity=1` when starting the plugin backend
```
jgbernalp, line 76: should this be `-log-level`?

PeterYurkovich: sure should be, I'll put a PR up to fix it later

PR #189, opened 2024-09-19
RHINENG-12334: filters for incidents

Fewwy:
Added functionality for filters 
I may need to fix things in the future when real incident data is present in the Prometheus database.
I added a JSdoc for the filter function, I found it much more useful than any typescript interfaces and other stuff, because it provides highlights and allows you to explain how the function works + provide an example

### DIFF
--- a/src/components/Incidents/IncidentsPage.jsx
+++ b/src/components/Incidents/IncidentsPage.jsx
@@ -49,7 +49,7 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
       { id: 'informative', title: t('Informative') },
       { id: 'inactive', title: t('Inactive') },
     ],
-    type: 'informative',
+    type: 'incident-type',
   });
   const [staticData, filteredData, onFilterChange] = useListPageFilter(incidentsData, [
     incidentTypeFilter(t),
--- a/src/components/Incidents/IncidentsPage.jsx
+++ b/src/components/Incidents/IncidentsPage.jsx
@@ -8,7 +8,12 @@ import {
 } from '@openshift-console/dynamic-plugin-sdk';
 import { parsePrometheusDuration } from '../console/utils/datetime';
 import { getPrometheusURL } from '../console/graphs/helpers';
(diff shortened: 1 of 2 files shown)


src/components/Incidents/IncidentsPage.jsx
```diff
@@ -36,19 +41,20 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
...
+    filter: (filter, incident) => filterIncident(incident, filter),
     filterGroupName: t('Incident type'),
     isMatch: () => console.log('here should be filter state'),
     items: [
       { id: 'long-standing', title: t('Long standing') },
       { id: 'informative', title: t('Informative') },
       { id: 'inactive', title: t('Inactive') },
     ],
-    type: 'name',
+    type: 'informative',
```
jgbernalp, line 52: I think this type should reflect the filter grouping in this case should it be "incident-status" or something similar?

src/components/Incidents/utils.js
```diff
@@ -11,8 +12,6 @@ export function groupAndDeduplicate(objects, keyType) {
       key = obj.metric.alertname + obj.metric.namespace;
     } else if (keyType === 'group_id') {
       key = obj.metric.group_id;
-    } else {
```
jgbernalp, line 14: If this is removed, we would create a new entry with the undefined key for the groupedObjects Map

Fewwy: This is not an issue, there are 2 places where this function is called. processAlertTimestamps and processIncidentTimestamps and in both cases we directly pass the string parameter.

src/components/Incidents/IncidentsPage.jsx
Fewwy: I will update the type

src/components/Incidents/utils.js
jgbernalp: In that case we can define a type for the `keyType: 'alertname+namespace' | 'group_id'` to make this function more robust at least in compilation time as TypeScript will check that the correct values are passed as params

PR #190, opened 2024-09-19
add jsdoc for 2 utils funcitons

Fewwy:
I added jsdoc for 2 functions so I could have highlights that describes them + it will be easier to write tests for them in the future

### DIFF
--- a/src/components/Incidents/utils.js
+++ b/src/components/Incidents/utils.js
@@ -1,7 +1,29 @@
+/* eslint-disable max-len */
 import global_danger_color_100 from '@patternfly/react-tokens/dist/esm/global_danger_color_100';
 import global_info_color_100 from '@patternfly/react-tokens/dist/esm/global_info_color_100';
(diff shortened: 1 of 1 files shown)


src/components/Incidents/utils.js
```diff
@@ -1,7 +1,29 @@
...
+ *
+ * @description
+ * This function processes an array of objects by grouping them according to a key derived from `keyType`.
+ * - If `keyType` is `'alertname+namespace'`, objects are grouped by concatenating the `alertname` and `namespace` fields in the `metric` object.
+ * - If `keyType` is `'group_id'`, objects are grouped by the `group_id` field in the `metric` object.
+ *
+ * For each group, the values from all grouped objects are merged, and duplicates are removed based on the serialized form of each value.
+ *
+ * The result is an array of objects, each containing the `metric` from one of the grouped objects and the deduplicated `values`.
+ */
```
jgbernalp, line 26: This is ok for documentation but it would be better to use the actual types on the parameters so TypeScript can check the usages

PR #192, opened 2024-09-21
incidents-filter-matcher

Fewwy:
image

1. Set up matcher function so it could provide label numbers for filter dropdown
2. I refactored the processIncidentTimestamps a bit so it could assign boolean values correctly

In the next PR I will look into refactoring filterIncident function so I could utilize these boolean values for filtering and avoid doing double filtering

### DIFF
--- a/src/components/Incidents/IncidentsPage.jsx
+++ b/src/components/Incidents/IncidentsPage.jsx
@@ -26,7 +26,7 @@ import {
 } from '@patternfly/react-core';
 import { Helmet } from 'react-helmet';
 import { useBoolean } from '../hooks/useBoolean';
-import * as _ from 'lodash-es';
+import some from 'lodash-es/some';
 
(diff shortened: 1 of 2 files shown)


src/components/Incidents/IncidentsPage.jsx
```diff
@@ -35,15 +36,17 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
   const [alertsData, setAlertsData] = React.useState([]);
   const [incidentsData, setIncidentsData] = React.useState([]);
   const [isOpen, setIsOpen, , setClosed] = useBoolean(false);
+
+  const incidentsMatcher = (incident) => _.some(incident, 'informative');
```
jgbernalp, line 40: If the property is called state, then it should be `_.some(incident, {"state":"informative"});`

jgbernalp (approved):

PR #198, opened 2024-09-25
feat(RHINENG-10902): Incidents table implementation

Fewwy:
I did the following things in this PR:

1. Added basic VirtualizedTable component, I will continue working with it in the next PRs. The reason for that is that I had to update the Alerts API request to combine/merge data on incidents and alerts.
2. Separated query-generating functions into api.js file. I want to leave utils.js for smaller functions unrelated to the API.
3. Added a function collectIncidentsDataForApiQuery, which collects the data of the incident and creates an array of objects with values.
4. Output of collectIncidentsDataForApiQuery is used in the createAlertsQuery function. This function is responsible for mapping over the input and creating API queries for the Alerts endpoint. The reason is the following: Ivan told me that some of the incidents could be missing src_alertname/src_namespace/src_severity/src_name. So we have to account for one of these values is missing. 
In addition to that, we're adding "layer" and "component" from the Incidents. These values are missing in the Alerts endpoint, but they are required for the table, so we have to "merge" data together.

I have a meeting with Ivan on Tuesday where we will discuss if this is the way how we should gather the data for the table.

5. Added a function groupAlertsForTable, which grabs the alert data fetched by API request and creates table rows based on the "component" property.
6. Updated IncidentsTableRow component to consume the new data and create rows with component names, and icons and display if the alert is firing.

image

7. Small updates
Updated the alerts chart tooltip
Updated the createAlertsChartBars function to color alert bars depending on the severity.
Added consts.js file
I moved the Dropdown items for Days filter to the consts.js so it won't take too much space in IncidentsPage.jsx

### DIFF
--- a/src/components/Incidents/IncidentsPage.jsx
+++ b/src/components/Incidents/IncidentsPage.jsx
@@ -11,6 +11,7 @@ import { parsePrometheusDuration } from '../console/utils/datetime';
 import { getPrometheusURL } from '../console/graphs/helpers';
 import {
   collectSeverityAndAlertnames,
+  createAlertsQuery,
   filterIncident,
   getIncidentsTimeRanges,
   processAlertTimestamps,
@@ -91,7 +92,7 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
                 endpoint: PrometheusEndpoint.QUERY_RANGE,
                 endTime: range.endTime,
                 namespace,
-                query: 'ALERTS',
+                query: createAlertsQuery(alertValuePairs),
                 samples: 24,
                 timespan: range.duration - 1,
               },
@@ -111,7 +112,7 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
           console.log(err);
         });
     })();
-  }, [span]);
+  }, [alertValuePairs]);
 
   React.useEffect(() => {
     setalertValuePairs(collectSeverityAndAlertnames(incidentsData));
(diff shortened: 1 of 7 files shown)


src/components/Incidents/IncidentsPage.jsx
```diff
@@ -35,6 +39,7 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
   const [span, setSpan] = React.useState(parsePrometheusDuration('7d'));
   const [alertsData, setAlertsData] = React.useState([]);
   const [incidentsData, setIncidentsData] = React.useState([]);
+  const [alertValuePairs, setalertValuePairs] = React.useState([]);
```
jgbernalp, line 42: nit
```suggestion
  const [alertValuePairs, setAlertValuePairs] = React.useState([]);
```

src/components/Incidents/IncidentsPage.jsx
```diff
@@ -108,6 +113,10 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
     })();
   }, [span]);
 
+  React.useEffect(() => {
+    setalertValuePairs(collectSeverityAndAlertnames(incidentsData));
```
jgbernalp, line 117: Should be this called when the incidents finished loading?

Fewwy: Yes, I updated this in the latest commit.
So the scenario is following - We request the incidents, get "src_alertname" and "scr_severity" from them, then create custom unique pairs of values and then make new query that will look like
ALERTS{severity="smth",alertname="smth"} or ALERTS etc. 
That will be done for each custom combination of values. This is how we can request Alerts that match the Incidents.

PR #200, opened 2024-09-30
OU-550: Use npm instead of yarn

jgbernalp:


PeterYurkovich:
fyi, `yarn lint` is hardcoded into the release repo, so there will need to be a change there to fix that: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/monitoring-plugin/openshift-monitoring-plugin-release-4.17.yaml#L36.

jgbernalp:
Thx @PeterYurkovich added the fix for the CI https://github.com/openshift/release/pull/57306

PR #201, opened 2024-09-30
NO-JIRA: Make command for building images with MacOS

zhuje:


PeterYurkovich:
I do something similar in my dotfiles, you can just check the $OSTYPE variable https://github.com/PeterYurkovich/dotfiles/blob/main/common/.commonrc#L1

PR #202, opened 2024-10-02
OU-507: Add ACM alerting

PeterYurkovich:
This PR is still a major work in progress, I'm putting it up for others to track progress if they would like. The helm chart is able to deploy alongside the default `monitoring-plugin`. All backend work should be completed at this point, so reviews on that front.

The helm chart is able to deploy alongside the default `monitoring-plugin`. The extension points look to be mostly correct, however the page in ACM is currently blank (and is the next point of work). Wholistic perspective updates have been made on the frontend, and although they appear to work from some simple clicking around, more testing is required for them.


PeterYurkovich:
Most recent commit looks to duplicate the `observe` store which is used by the monitoring-plugin. The issue is that the store is located in the `openshift/console` codebase, which means any changes will only be available in 4.18. The codebase also has hardcoded items like [this](https://github.com/openshift/console/blob/b1259c14e30bb91f78ba0892c2bb6b91c4ead8b0/frontend/public/reducers/observe.ts#L137) which prevent the usage of a generic "perspective" change by locking to only dev and admin.


Due to these issues we need to bring the store changes into the plugin instead. These updates should also lay the groundwork for migrating the store from the console codebase into this one, by updating the functions within the monitoring-plugin with the correct perspective code and by routing actions through functions, which can then be internally changed to point to the same store or removed later. While it should make some parts of transitioning easier, it could also make other parts more difficult, by fragmenting the store into two locations, leading to a future which requires recombining them somehow.

Dockerfile.art
```diff
@@ -43,5 +43,6 @@ USER 1001
 
 COPY --from=web-builder /usr/src/app/web/dist /opt/app-root/web/dist
 COPY --from=go-builder $REMOTE_SOURCES_DIR/cachito-gomod-with-deps/app/plugin-backend /opt/app-root
+COPY config/ /opt/app-root/config
```
jgbernalp, line 46: This should be copied from the web builder step

Dockerfile.ci
```diff
@@ -50,5 +50,6 @@ USER 1001
 
 COPY --from=web-builder /opt/app-root/src/web/dist /opt/app-root/web/dist
 COPY --from=go-builder /opt/app-root/plugin-backend /opt/app-root/plugin-backend
+COPY config/ /opt/app-root/config
```
jgbernalp, line 53: This should be copied from the web builder step

pkg/server.go
```diff
@@ -102,30 +143,63 @@ func Start(cfg *Config) {
...
 	} else {
-		log.Infof("listening on http://:%d", cfg.Port)
+		log.Infof("listening on http. port: %d", cfg.Port)
 		logrus.SetLevel(logrusLevel)
 		panic(httpServer.ListenAndServe())
 	}
 }
 
-func setupRoutes(cfg *Config) (*mux.Router, *PluginConfig) {
+func setupRoutes(cfg *Config, acmMode bool) (*mux.Router, *PluginConfig) {
```
jgbernalp, line 162: does the cfg hold the features already, we will have more features like incidents

web/src/components/alerting.tsx
```diff
@@ -510,11 +521,10 @@ const SilencesDetailsPage_: React.FC<RouteComponentProps<{ id: string }>> = ({ m
...
             <BreadcrumbItem>
-              <Link
-                className="pf-c-breadcrumb__link"
-                to={isDev ? `/dev-monitoring/ns/${namespace}/alerts` : '/monitoring/silences'}
-              >
-                {isDev ? t('Alerts') : t('Silences')}
+              <Link className="pf-c-breadcrumb__link" to={getAlertsUrl(perspective, namespace)}>
+                {t('Alerts')}
+                {/* TODO: this setup is wrong, need to determine behavior for each perspective.*/}
+                {/* I think it should be linking to silences pages always, but unclear */}
```
jgbernalp, line 527: yes it should be silences, the ACM case will mimic the admin perspective UX

Dockerfile.art
PeterYurkovich: Is there a reason to first copy it into the web-builder and then into the final image, rather than just putting it directly into the final image?

pkg/server.go
PeterYurkovich: It had made the logic a bit easier, but I agree that adding it doesn't scale too well as we add more features. I'll remove it

Dockerfile.art
jgbernalp: no specific reason, we can leave it as it is

Dockerfile.acm
```diff
@@ -0,0 +1,49 @@
...
+
+WORKDIR /opt/app-root
+
+ENV HUSKY=0
+
+COPY Makefile Makefile
+COPY web/ web/
+
+# Replace the plugin name in the package.json for the custom build actions
+RUN sed -i 's/"name": "monitoring-plugin",/"name": "monitoring-console-plugin",/g' web/package.json
```
jgbernalp, line 13: should we add these under `/scripts`, WDYT?

Dockerfile.ci
jgbernalp, line 1: Is this file used by the openshift/release bot?

Dockerfile.acm
PeterYurkovich: Makes sense to me, especially since it's duplicated in the deploy-acm.sh script

config/clear-extensions.patch.json
jgbernalp, line 1: I'm not sure I understand what clear-extensions mean on this patch. Is it to add the monitoring reducer in the case there is none provided by the console? if so does it make sense to include it always and use only this one and not the one in the console?

scripts/build-image.sh
```diff
@@ -6,6 +6,7 @@ PREFER_PODMAN="${PREFER_PODMAN:-1}"
 PUSH="${PUSH:-0}"
 TAG="${TAG:-v1.0.0}"
 REGISTRY_ORG="${REGISTRY_ORG:-openshift-observability-ui}"
+DOCKER_FILE_NAME="${DOCKER_FILE_NAME:-Dockerfile.local}"
```
jgbernalp, line 9:
```suggestion
DOCKER_FILE_NAME="${DOCKER_FILE_NAME:-Dockerfile.dev}"
```

web/src/components/alerting/AlertUtils.tsx
```diff
@@ -51,11 +51,12 @@ export const alertingRuleSource = (rule: Rule): AlertSource | string => {
 };
 
 export const alertSource = (alert: Alert): AlertSource | string => alertingRuleSource(alert.rule);
+export const alertCluster = (alert: Alert): string => alert.labels?.cluster ?? '';
 
 export const SilencesNotLoadedWarning: React.FC<{ silencesLoadError: any }> = ({
   silencesLoadError,
 }) => {
-  const { t } = useTranslation('plugin__monitoring-plugin');
+  const { t } = useTranslation(process.env.NAMESPACE);
```
jgbernalp, line 59: should we use
```suggestion
  const { t } = useTranslation(process.env.I18N_NAMESPACE);
```
?

web/src/components/alerting/AlertsPage.tsx
```diff
@@ -155,21 +164,38 @@ const AlertsPage_: React.FC<AlertsPageProps> = () => {
...
-    [t],
-  );
+    ];
+
+    if (perspective === 'acm') {
+      cols.splice(-1, 0, {
+        id: 'cluster',
+        props: { className: tableAlertClasses[4] },
+        sort: 'labels.cluster',
+        title: t('Cluster'),
```
jgbernalp, line 177: maybe not for this PR, but this should probably be a link to the alert in the spoke

Dockerfile.ci
PeterYurkovich: It is an unused dockerfile

config/clear-extensions.patch.json
PeterYurkovich: The clear extensions main purpose is to remove all other extensions from usage. This will prevent duplicate tabs in the developer perspective and the like. However, the store will be likely be used by multiple features, so it made sense to leave it after clearing the rest out.

The primary purpose for adding an additional store is to change the reducer logic present in the console. The options were to either:
1. Duplicate each reducer function we need to change with slightly different names so that the ones in the monitoring-plugin are used rather than the ones in the console.
2. Duplicate the store.

Since we are already planning to move the store out of the console codebase I thought it would make sense to go ahead and duplicate the store instead. I could understand going ahead and doing the full migration now, but it seemed like a large commitment that wasn't needed, and would need updates to the console (for things like the alerts rules page). 

I think it would be best to add the store now, and then make the migration to the new monitoring-plugin exclusive store once we have moved the rest of the plugin code out of the console codebase.

PeterYurkovich:
Squashed down older commits, leaving behind the ones we have had discussions on this past week

jgbernalp:
@PeterYurkovich  This LGTM, should we now ask QE to check the regular and ACM images for regressions?

PeterYurkovich:
As per our discussion, we will be holding off on the ACM testing until the ObO PR is ready. 

@Tai-RedHat Could QE perform regression testing on this PR?

Tai-RedHat:
> As per our discussion, we will be holding off on the ACM testing until the ObO PR is ready.
> 
> @Tai-RedHat Could QE perform regression testing on the normal CMO deployed plugin using this PR?

test this pr with cluster-bot, normal CMO deployed plugin worked well.
`launch 4.18,openshift/monitoring-plugin#202 gcp`

PR #205, opened 2024-10-08
RHINENG-12637: add onclick to incidents chart

Fewwy:
https://issues.redhat.com/browse/RHINENG-12637

1. Added size observer to both charts
2. Added on click function to the Incidents chart (it is not required for the Alerts chart)
when the user clicks on a chart bar it makes the bars that aren't chosen 30% more transparent
3. I moved files that are related to processing Incidents and Alerts into 2 separate files and added JSdoc to them because utils.js was getting too big. That will allow us to find the functions related to Incidents and Alerts faster
4. I had to update the query builder for the Alerts because we're requesting them and merging them with values from Incidents. 
I added layerList and componentList array to collect all the possible layers and components for the incident. Then I built a long query that requests every combination of the group_id (it's static because Incident has 1 group id) + variety of layer/component values.
5. I had to refactor the way I group Incidents and Alerts together to make sure that we're not losing any Alerts data and that they are grouped up differently for either the Alerts chart or the Alerts table.
6. I will update the default Alerts chart and default table view when Fodal provides me a mock. It will be done in a separate PR.
Future improvements:
8. I am not sure that I handle the logic for onClick and Loading/Rerendering the table correctly. I assume it could be optimized more, then it would be faster.
9. I have to make a custom hook that would handle fetch for alerts/incidents, and then I can move that code out of the IncidentsPage component.

Small fix:
I updated the way we show the date for alerts start/end firing in the expanded details because it was incorrect before.
image


src/components/Incidents/AlertsChart/AlertsChart.jsx
```diff
@@ -25,84 +24,100 @@ const AlertsChart = ({ alertsData, chartDays }) => {
...
+              legendPosition="bottom-left"
+              //this should be always less than the container height
+              height={350}
+              padding={{
+                bottom: 75, // Adjusted to accommodate legend
+                left: 50,
+                right: 25, // Adjusted to accommodate tooltip
+                top: 0,
+              }}
+              //TODO: WIDTH SHOULD BE AUTOMATICALLY ADJUSTED
```
jgbernalp, line 89:
```suggestion
```

src/components/Incidents/IncidentsHeader/IncidentsHeader.jsx
```diff
@@ -2,10 +2,21 @@ import * as React from 'react';
 import IncidentsChart from '../IncidentsChart/IncidentsChart';
 import AlertsChart from '../AlertsChart/AlertsChart';
 
-export const IncidentsHeader = ({ alertsData, incidentsData, chartDays }) => {
+export const IncidentsHeader = ({
+  alertsData,
+  incidentsData,
+  chartDays,
+  chooseIncident,
```
jgbernalp, line 9: I think this might be better named as `selectedIncident`

src/components/Incidents/IncidentsPage.jsx
```diff
@@ -141,6 +137,38 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
     })();
   }, [span]);
 
+  React.useEffect(() => {
+    (async () => {
```
jgbernalp, line 141: we might not need async here as the only await used is inside the `timeRanges.map`

src/components/Incidents/IncidentsPage.jsx
```diff
@@ -176,6 +204,8 @@ const IncidentsPage = ({ customDataSource, namespace }) => {
             alertsData={alertsData}
             incidentsData={filteredData}
             chartDays={timeRanges.length}
+            chooseIncident={setChooseIncident}
+            tableLoaded={setAlertsAreLoading}
```
jgbernalp, line 208: if these are function props it might be more readable to use:
```suggestion
            onIncidentSelect={setChooseIncident}
            onTableLoaded={setAlertsAreLoading}
```

src/components/Incidents/api.js
```diff
@@ -48,56 +72,50 @@ export const collectIncidentsDataForApiQuery = (objects) => {
...
  */
 export const createAlertsQuery = (groupedAlertsValues) => {
-  // Map through the grouped alerts to create individual alert strings for the query
   const alertsQuery = groupedAlertsValues
     .map((query) => {
-      const alertParts = ['alertname', 'namespace', 'name', 'severity', 'alertstate']
-        .filter((key) => query[key]) // Only include keys that are present in the query object
-        .map((key) => `${key}="${query[key]}"`)
-        .join(', ');
+      // Generate query for each combination of component and layer
```
jgbernalp, line 108: This query seems to be very expensive to execute. Can we execute queries lazily only per component/layer?

src/components/Incidents/IncidentsChart/IncidentsChart.jsx
```diff
@@ -6,103 +6,147 @@ import {
...
+                tickValues={dateValues}
+              />
+              <ChartGroup horizontal>
+                {chartData.map((bar, index) => {
+                  return (
+                    //we have several arrays and for each array we make a ChartBar
+                    <ChartBar
+                      name={bar.group_id}
+                      data={bar}
+                      key={index}
```
jgbernalp, line 128: is there a better index we can use as a key?, this will prevent unnecessary re renders

src/components/Incidents/IncidentsChart/IncidentsChart.jsx
```diff
@@ -6,103 +6,147 @@ import {
...
+  }, []);
   React.useEffect(() => {
     setIsLoading(false);
     setChartData(incidentsData.map((incident) => createIncidentsChartBars(incident)));
   }, [incidentsData]);
+  const dateValues = generateDateArray(chartDays);
+
+  const isHidden = (group_id) => hiddenSeries !== null && hiddenSeries !== group_id;
+  const clickHandler = (data, datum) => {
+    tableLoaded(true);
```
jgbernalp, line 43: not sure why this needs to change here. This is probably changing the state of the parent component forcing a re-render of this component.

src/components/Incidents/IncidentsChart/IncidentsChart.jsx
```diff
@@ -6,103 +6,147 @@ import {
...
+    setHiddenSeries(datum.datum.group_id === hiddenSeries ? null : datum.datum.group_id);
+    chooseIncident(datum.datum.group_id);
```
jgbernalp, line 45: Same, here. It might be better to keep an internal state of the chosenIncident, as we don't want the parent to re render every time.

src/components/Incidents/IncidentsChart/IncidentsChart.jsx
Fewwy: I am not using name prop, so I can replace index with bar.group_id

src/components/Incidents/IncidentsHeader/IncidentsHeader.jsx
jgbernalp: ok then is an action, in that case it might be better `onSelectIncident`

Fewwy: I adjusted it according to the other comment

src/components/Incidents/IncidentsChart/IncidentsChart.jsx
```diff
@@ -6,103 +6,144 @@ import {
...
+  }, []);
   React.useEffect(() => {
     setIsLoading(false);
     setChartData(incidentsData.map((incident) => createIncidentsChartBars(incident)));
   }, [incidentsData]);
+  const dateValues = generateDateArray(chartDays);
+
+  const isHidden = (group_id) => hiddenSeries !== null && hiddenSeries !== group_id;
+  const clickHandler = (data, datum) => {
+    tableIsLoading(true);
```
jgbernalp, line 43: Do we need to set that the table is loading from this children component?. It seems the parent component is the one requesting the table when the selected incident changes.

PR #233, opened 2024-10-24
RHINENG-10907: Filter to toolbar refactor

Fewwy:
https://issues.redhat.com/browse/RHINENG-10907
1) removed openshift filter
2) added toolbar filter + dropdown
3) connected all of the filters to the state management and data processing
4) added hide graph button that hides it
image

### DIFF
--- a/web/src/components/Incidents/IncidentsPage.jsx
+++ b/web/src/components/Incidents/IncidentsPage.jsx
@@ -2,34 +2,22 @@
 import * as React from 'react';
 import { IncidentsHeader } from './IncidentsHeader/IncidentsHeader';
 import { useSafeFetch } from '../console/utils/safe-fetch-hook';
-import {
-  ListPageFilter,
-  PrometheusEndpoint,
-  useListPageFilter,
-} from '@openshift-console/dynamic-plugin-sdk';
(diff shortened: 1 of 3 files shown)


web/src/components/Incidents/IncidentsPage.jsx
```diff
@@ -2,68 +2,109 @@
...
 import { IncidentsTable } from './IncidentsTable';
 import { getIncidentsTimeRanges, processIncidents } from './processIncidents';
-import { filterIncident } from './utils';
+import {
+  filterIncident,
+  onDeleteGroupIncidentFilterChip,
+  onDeleteIncidentFilterChip,
+} from './utils';
 import { groupAlertsForTable, processAlerts } from './processAlerts';
+import { DropdownToggle, Dropdown } from '@patternfly/react-core/deprecated';
```
PeterYurkovich, line 32: Could you name these as deprecated like [this](https://github.com/openshift/monitoring-plugin/blob/aae8da3647e2637132e8f9f1475ba83558817243/web/src/components/silence-form.tsx#L16)

PeterYurkovich (changes requested):

PR #236, opened 2024-10-26
DRAFT: Connecting incidents to redux

Fewwy:


web/src/components/Incidents/IncidentsPage.jsx
```diff
@@ -169,6 +170,7 @@ const IncidentsPage = ({ customDataSource, namespace = '#ALL_NS#' }) => {
           setIncidentsData(processIncidents(aggregatedData));
           setFilteredData(filterIncident(filters, processIncidents(aggregatedData)));
           setIncidentsAreLoading(false);
+          dispatch(setIncidentsNavFilters('incidents', filters));
```
jgbernalp, line 173: you probably don't need the key as this can be set in the reducer
```suggestion
          dispatch(setIncidentsNavFilters(filters));
```

web/src/actions/observe.ts
```diff
@@ -134,6 +136,12 @@ export const queryBrowserToggleSeries = (index: number, labels: { [key: string]:
 
 export const setAlertCount = (alertCount) => action(ActionType.SetAlertCount, { alertCount });
 
+export const setIncidentsNavFilters = (key: AlertingKey, incidents) =>
+  action(ActionType.SetIncidentsNavFilters, {
+    key,
```
jgbernalp, line 141:
```suggestion
    key: "incidents",
```

web/src/actions/observe.ts
```diff
@@ -134,6 +136,12 @@ export const queryBrowserToggleSeries = (index: number, labels: { [key: string]:
...
+    incidents,
+  });
```
jgbernalp, line 143:
```suggestion
  action(ActionType.SetIncidentsNavFilters, { incidents });
```

Fewwy:
@jgbernalp I changed the code as you proposed but the 
  `const filtersInRedux = useSelector(({ observe }) => console.log(observe));`
is undefined
If I console log the oberve in selector I get this
image

It seems that filters are not added to the root and I don't see any other way of how should I add them there

PR #256, opened 2024-11-13
NO-JIRA: incidents feature squashed into 1 commit

Fewwy:


juzhao:
/cc: @Tai-RedHat please help to do regression test and file issue if it's there is

Tai-RedHat:
PR tested with cluster-bot: able to see `incidents` tab, other monitoring-plugin components work properly.
BTW, do we plan to add `incidents` tab to the developer console?

PR #260, opened 2024-11-18
OCPBUGS-44447: Don't set property on undefined variable

PeterYurkovich:
Don't try to set the notificationAlerts.data until it has been initialized. 

Bug is able to be reproduced when refreshing the alerting page or opening the alerting page in a new tab.


PeterYurkovich:
QE: @juzhao could you or someone else on your team please take a look?

Reproduction: Issue occurs on refreshing the alerting page or opening the alerting page in a new tab. Solution should prevent this issue and continue to allow the notification bell to still display the number of alerts.

PR #263, opened 2024-11-20
NO-JIRA: updated api request to remove prom tenancy issue

Fewwy:
This should fix the issue that caused request to prometheus-tenancy endpoint instead of regular endpoint

### DIFF
--- a/web/src/components/Incidents/api.js
+++ b/web/src/components/Incidents/api.js
@@ -61,24 +61,19 @@ export const createAlertsQuery = (groupedAlertsValues) => {
   return alertsQuery;
 };
 
-export const fetchDataForIncidentsAndAlerts = (
-  fetch,
-  range,
-  namespace,
-  customDataSource,
-  customQuery,
-) => {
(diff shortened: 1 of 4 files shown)


web/src/components/Incidents/api.js
```diff
@@ -61,24 +61,19 @@ export const createAlertsQuery = (groupedAlertsValues) => {
...
         endTime: range.endTime,
-        namespace,
+        namespace: '',
         query: customQuery,
         samples: 24,
         timespan: range.duration - 1,
       },
-      customDataSource?.basePath,
+      perspective,
+      '/api/prometheus',
```
PeterYurkovich, line 76: This parameter isn't needed as it is set in the function to the non-tenancy path by default. You can see an example [here](https://github.com/openshift/monitoring-plugin/blob/main/web/src/components/hooks/useRulesAlertsPoller.ts#L30)

PeterYurkovich (changes requested):

PR #265, opened 2024-11-21
NO-JIRA: only 1 alert in table fix

Fewwy:
This should fix issue with only 1 alert presented when user click on incident

image


PeterYurkovich:
/rename NO-JIRA: only 1 alert in table fix

PeterYurkovich:
/retitle NO-JIRA: only 1 alert in table fix

PR #266, opened 2024-11-21
OU-432: Merge Perses dashboards with current dashboards list 

zhuje:
### Description
On the dashboard page, list [Perses](https://github.com/perses/perses) dashboards in the dropdown. 
This PR only lists the dashboards in the dropdown; the dashboard is not rendered when you click on the dropdown item. 

OU-539_demo_list_perses_dashboards_ocp

**Figure 1.** Perses dashboards are listed in the dashboard page dropdown  (e.g., "testproject / dashboard1" and "testproject / dashboard2"). The dashboards are displayed as <"project name"> / <"dashboard name">. The tag, by default, is "perses."

### Related Issue 
https://issues.redhat.com/browse/OU-432

### Testing 
This PR was only tested locally during development. To setup the test: 
1. Launch a perses container `docker run --name perses -d -p 127.0.0.1:8080:8080 persesdev/perses`. See https://github.com/perses/perses?tab=readme-ov-file#docker-images for information. 
2. Navigate to `http://localhost:8080`, which will take you to the local Perses instance. Create a project and a dashboard for it. 
3. Start the monitoring-plugin (`make start-frontend` and in a separate terminal `make start-console`). `make start-console` is proxying localhost:8080 (our local perses instance). 
4. There is a bug in the OCP console that prevents the Dashboard page from fetching the OCP dashboards specs -- to circumvent (with a hack-y solution) I've created a mock API response that serves the specs in this branch which you can merge in to the PR : https://github.com/zhuje/monitoring-plugin/tree/mock_monitoring_dashboard_config.   
5. Navigate to `http://localhost:9000` (our local OCP instance). Go to Observe > Dashboards > Dropdown, the dashboards you created in the local perses instance should now be listed in the dropdown. 

https://github.com/user-attachments/assets/7f6258e8-62de-46dc-8bd1-c7c598d628ea
**Figure 2**. Video demo of testing (~1:45 minutes). Shows perses dashboards are listed in the Dashboard page dropdown.

### DIFF
--- a/web/src/components/dashboards/index.tsx
+++ b/web/src/components/dashboards/index.tsx
@@ -81,6 +81,7 @@ import {
 import KebabDropdown from '../kebab-dropdown';
 import { MonitoringState } from '../../reducers/observe';
 import { DropDownPollInterval } from '../dropdown-poll-interval';
+import { usePerses } from './usePerses';
 
 const intervalVariableRegExps = ['__interval', '__rate_interval', '__auto_interval_[a-z]+'];
 
@@ -904,6 +905,12 @@ const MonitoringDashboardsPage_: React.FC<MonitoringDashboardsPageProps> = ({ hi
   const { perspective } = usePerspective();
   const [board, setBoard] = React.useState<string>();
   const [boards, isLoading, error] = useFetchDashboards(namespace);
+  const { getPersesDashboards, dashboardsData: persesDashboards } = usePerses();
+
+  // Called only once on mount
+  React.useEffect(() => {
+    getPersesDashboards();
+  }, [getPersesDashboards]);
 
   // Clear queries on unmount
   React.useEffect(() => () => dispatch(queryBrowserDeleteAllQueries()), [dispatch]);
@@ -918,14 +925,24 @@ const MonitoringDashboardsPage_: React.FC<MonitoringDashboardsPageProps> = ({ hi
     [perspective, dispatch],
   );
 
-  const boardItems = React.useMemo(
-    () =>
-      _.mapValues(_.mapKeys(boards, 'name'), (b, name) => ({
-        tags: b.data?.tags,
-        title: b.data?.title ?? name,
-      })),
-    [boards],
-  );
+  const boardItems = React.useMemo(() => {
+    const ocpBoardItems = _.mapValues(_.mapKeys(boards, 'name'), (b, name) => ({
+      tags: b.data?.tags,
+      title: b.data?.title ?? name,
+    }));
+
+    if (persesDashboards) {
+      const persesKeys = _.mapKeys(persesDashboards, function (item) {
+        return item?.metadata?.name;
+      });
+      const persesBoardItems = _.mapValues(persesKeys, (b, name) => ({
+        tags: ['perses'],
+        title: `${b.metadata?.project} / ${b.metadata?.name}` ?? name,
+      }));
+      return { ...persesBoardItems, ...ocpBoardItems };
+    }
+    return ocpBoardItems;
+  }, [boards, persesDashboards]);
 
   const changeBoard = React.useCallback(
     (newBoard: string) => {
--- a/web/src/components/dashboards/perses-client.ts
+++ b/web/src/components/dashboards/perses-client.ts
@@ -1,0 +1,21 @@
+import { cancellableFetch, CancellableFetch } from '../cancellable-fetch';
+
+const baseURL = '/api/proxy/plugin/monitoring-console-plugin/perses';
+
+export type PersesDashboardMetadata = {
+  kind: string; // kind: Dashboards
+  metadata: {
+    name: string; // name of dashboard
+    created: string;
+    updated: string;
+    version: string;
+    project: string; // name of project that holds the dashboard
+  };
+};
+
+export const fetchPersesDashboardsMetadata = (): CancellableFetch<PersesDashboardMetadata[]> => {
+  const listDashboardsMetadata = '/api/v1/dashboards?metadata_only=true';
+  const persesURL = `${baseURL}${listDashboardsMetadata}`;
+
+  return cancellableFetch<PersesDashboardMetadata[]>(persesURL);
+};
--- a/web/src/components/dashboards/usePerses.ts
+++ b/web/src/components/dashboards/usePerses.ts
@@ -1,0 +1,99 @@
+import { PersesDashboardMetadata } from './perses-client';
+import { useCallback, useReducer, useRef } from 'react';
+import { fetchPersesDashboardsMetadata } from './perses-client';
+
+const isAbortError = (error: unknown): boolean =>
+  error instanceof Error && error.name === 'AbortError';
+
+type State = {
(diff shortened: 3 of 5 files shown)


pkg/proxy/proxy.go
```diff
@@ -28,13 +28,15 @@ type KindType string
...
 	ThanosQuerierKind KindType = "thanos-querier"
+	PersesKind        KindType = "perses"
 )
 
 type ProxyPort int
 
 const (
 	AlertmanagerPort  ProxyPort = 9444
 	ThanosQuerierPort ProxyPort = 9445
+	PersesPort        ProxyPort = 9446
```
PeterYurkovich, line 39: Make sure to update the port argument info in plugin-backend to include this port as reserved

jgbernalp:
As discussed in the tech sync, I don't think we need the proxy here as this will be handled by the proxy created in the ConsolePlugin CR

pkg/proxy/proxy.go
zhuje: I'm going to remove the proxy so the CR can handle it.

cmd/plugin-backend.go
```diff
@@ -11,17 +11,18 @@ import (
...
+	keyArg                 = flag.String("key", "", "private key file path to enable TLS (disabled by default)")
+	featuresArg            = flag.String("features", "", "enabled features, comma separated.\noptions: ['acm-alerting', 'incidents', 'dev-config']")
+	staticPathArg          = flag.String("static-path", "", "static files path to serve frontend (default: './web/dist')")
+	configPathArg          = flag.String("config-path", "", "config files path (default: './config')")
+	pluginConfigArg        = flag.String("plugin-config-path", "", "plugin yaml configuration")
+	logLevelArg            = flag.String("log-level", logrus.InfoLevel.String(), "verbosity of logs\noptions: ['panic', 'fatal', 'error', 'warn', 'info', 'debug', 'trace']\n'trace' level will log all incoming requests\n(default 'error')")
+	alertmanagerUrlArg     = flag.String("alertmanager", "", "alertmanager url to proxy to for acm mode")
+	thanosQuerierUrlArg    = flag.String("thanos-querier", "", "thanos querier url to proxy to for acm mode")
+	log                    = logrus.WithField("module", "main")
+	persesDashboardsUrlArg = flag.String("perses-dashboards", "", "perses API endpoint to fetch dashboards")
```
jgbernalp, line 25: I don't think we need this url here as is known from the CR plugin side proxy.

web/src/components/dashboards/useFetchPersesDashboards.ts
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import * as _ from 'lodash';
+import { useSafeFetch } from '../console/utils/safe-fetch-hook';
+import { useBoolean } from '../hooks/useBoolean';
+
+export const useFetchPersesDashboards = (): [[], boolean, string] => {
```
jgbernalp, line 6: I think it will be bette to abstract the Perses API in a `persesClient.ts` and then use this hook to consume the API. This will allow to extend it later easily. We should type the hook response so is easier to read.
```suggestion
export const usePerses = (): PersesHook => {
```

jgbernalp (changes requested):

fkargbo:
Looks good to me!

cmd/plugin-backend.go
zhuje: I removed this in the latest push.

web/src/components/dashboards/useFetchPersesDashboards.ts
zhuje: I refactored this into `persesClient.ts` in the latest push.

web/src/components/dashboards/index.tsx
```diff
@@ -905,6 +906,9 @@ const MonitoringDashboardsPage_: React.FC<MonitoringDashboardsPageProps> = ({ hi
   const [board, setBoard] = React.useState<string>();
   const [boards, isLoading, error] = useFetchDashboards(namespace);
 
+  const { getDashboards } = usePersesClient();
+  const [persesBoards] = getDashboards;
```
jgbernalp, line 910: This probably should be a function so it can be lazy executed.
```suggestion
  const persesBoards = getDashboards();
```

web/src/components/dashboards/perses-client.ts
jgbernalp, line 1: It would be better to leave the perses-client to abstract the perses API as functions. See:
https://github.com/openshift/logging-view-plugin/blob/main/web/src/loki-client.ts

And then build in another file a `usePerses` hook that consumes this client. This hook should expose functions and probably a state so the consumers can decide when to query or reuse a persistent state.

zhuje: I updated the files in the latest push. 
1. Added cancellable-fetch.ts ported from logging-view-plugin
2. perses-client.ts modified to follow loki-client.ts from logging-view-plugin
3. usePerses.ts created to follow useLogs.ts from logging-view-plugin
4. ...dashboards/index.ts modified to utilize usePerses() hook

web/src/components/dashboards/index.tsx
zhuje: Removed in the latest push.

web/src/components/dashboards/usePerses.ts
```diff
@@ -0,0 +1,99 @@
...
+        isLoadingDashboardsData: true,
+        dashboardsData: undefined,
+        dashboardsError: undefined,
+      };
+    default:
+      return state;
+  }
+};
+
+export const usePerses = () => {
```
PeterYurkovich, line 53: We should change this hook name to be something like "usePersesDashboardList", as there are other perses hook we will need to create in the future

web/src/components/dashboards/usePerses.ts
```diff
@@ -0,0 +1,99 @@
...
+  };
+
+  const [state, dispatch] = useReducer(reducer, initialState);
+
+  const { dashboardsData, isLoadingDashboardsData, dashboardsError } = state;
+
+  /**
+   * Lists perses dashboards meta data only (no dashboards specs)
+   */
+  const getPersesDashboards = useCallback(async () => {
```
PeterYurkovich, line 69: I think instead of having a function to manually retrigger the fetch, it might be better to instead just do polling like how the alerts and silences have

zhuje: This might be a separate JIRA ticket to enhance polling. I create the issue here: [OU-615](https://issues.redhat.com/browse/OU-615).

web/src/components/dashboards/usePerses.ts
zhuje: I'm following the structure set by https://github.com/openshift/logging-view-plugin/blob/1319d70979d33cfa936dae6890c7bb02d1fa961d/web/src/hooks/useLogs.ts#L4 per our meeting on Thursday. 

I believe the other Perses hooks we create in the feature will live within usePerses.

PR #268, opened 2024-11-22
OU-590: upgrade dynamic plugin sdk webpack to load correct patternfly dependency

jgbernalp:
This PR updates the dynamic-plugin-sdk-webpack. With this change the exposed modules that are not used in the extensions triggers a build error, causing the Docker images to fail. 

I moved the `plugin-manifest.json` to the config folder, similar to the logging view plugin. And added all the exposed modules in the `console-extensions.json`. This allows the plugin to build correctly but now the `plugin-manifest.json` needs to be loaded from the config folder to avoid injecting extensions that come from features such as incidents.


pkg/plugin_handler.go
```diff
@@ -13,7 +13,7 @@ import (
 var mlog = logrus.WithField("module", "manifest")
 
 func manifestHandler(cfg *Config) http.HandlerFunc {
-	baseManifestData, err := os.ReadFile(filepath.Join(cfg.StaticPath, "plugin-manifest.json"))
+	baseManifestData, err := os.ReadFile(filepath.Join(cfg.ConfigPath, "plugin-manifest.json"))
```
jgbernalp, line 16: This is the main change

PR #273, opened 2024-11-26
NO-JIRA: Upgrade incidents dropdown

jgbernalp:
this PR upgrades the incidents type dropdown filter to use PF5


Fewwy:
Are you sure this fix is working? I cannot test it on the main branch (because of all the broken commands), but when I add this code on top of the commit when we merge incidents into the main branch, it doesn't work. I see an infinite loading spinner.

PR #276, opened 2024-11-26
NO-JIRA: Fix feature flagged DX

PeterYurkovich:
This PR adds better DX for running the monitoring-plugin with features enabled while developing.

This doesn't exactly add hot reloading, but the page does refresh when changes are made to the frontend

### DIFF
--- a/config/dev-config.patch.json
+++ b/config/dev-config.patch.json
@@ -1,0 +1,13 @@
+[
+  {
+    "op": "add",
+    "path": "/extensions/0",
+    "value": {
+      "type": "console.redux-reducer",
+      "properties": {
+        "scope": "monitoring",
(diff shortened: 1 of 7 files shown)


config/dev-config.patch.json
jgbernalp, line 1: Should we use a different folder so dev patches aren't loaded in prod by mistake. Meaning when building the image the `dev-config` folder won't exist

PeterYurkovich: IMO, adding another folder will leave us more open to mistakes by adding the extension in 1 place but not both. In order to use the dev-patch we would need to manually pass in the `dev-config` parameter, which seems less likely to cause issues to me than having duplicate extension folders.

PR #279, opened 2024-12-03
RHINENG-14677: fix filters for incidents

Fewwy:
Fixed filters not working. You can find recordings of the filters fix in the ticket below
https://issues.redhat.com/browse/RHINENG-14677

### DIFF
--- a/web/src/components/Incidents/processIncidents.ts
+++ b/web/src/components/Incidents/processIncidents.ts
@@ -1,9 +1,4 @@
 /* eslint-disable max-len */
-const currentDate = new Date(); // Get the current date and time in UTC
-const currentDay = currentDate.getUTCDate();
-const currentMonth = currentDate.getUTCMonth();
-const currentYear = currentDate.getUTCFullYear();
-const currentHour = currentDate.getUTCHours();
 
 interface Metric {
   src_alertname: string;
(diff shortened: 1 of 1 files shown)


web/src/components/Incidents/processIncidents.ts
```diff
@@ -62,17 +57,15 @@ export function processIncidents(data: Incident[]): ProcessedIncident[] {
...
-        valueDate.getUTCFullYear() === currentYear &&
-        valueDate.getUTCHours() === currentHour
-      );
-    });
+    const currentDate = new Date(); // Current timestamp in milliseconds
+
+    // Calculate the time difference in milliseconds
+    const timeDifference = currentDate.valueOf() - lastDate * 1000;
+    const inactive = timeDifference < 10 * 60 * 1000;
+    const dayDifference = (lastDate - firstDate) / (60 * 60 * 24);
```
zhuje, line 65: Should this be (1000 * 60 * 60 * 24)?

web/src/components/Incidents/processIncidents.ts
```diff
@@ -62,17 +57,15 @@ export function processIncidents(data: Incident[]): ProcessedIncident[] {
...
-        valueDate.getUTCMonth() === currentMonth &&
-        valueDate.getUTCFullYear() === currentYear &&
-        valueDate.getUTCHours() === currentHour
-      );
-    });
+    const currentDate = new Date(); // Current timestamp in milliseconds
+
+    // Calculate the time difference in milliseconds
+    const timeDifference = currentDate.valueOf() - lastDate * 1000;
+    const inactive = timeDifference < 10 * 60 * 1000;
```
zhuje, line 64: So the incident is considered inactive in 10 * 60 * 1000 = 600,000 ms = 10 minutes? Is that correct?

Fewwy: Yes

web/src/components/Incidents/processIncidents.ts
Fewwy: Yes

Fewwy: No, the last date and first date are already converted beforehand. They

zhuje: Sounds good.

PR #280, opened 2024-12-03
RHINENG-14585: Remove pending alerts from processing and add another value to the unique key

Fewwy:
https://issues.redhat.com/browse/RHINENG-14585

### DIFF
--- a/web/src/components/Incidents/processAlerts.js
+++ b/web/src/components/Incidents/processAlerts.js
@@ -27,9 +27,11 @@
  * // Returns an array where the two alerts are grouped together with deduplicated values.
  */
 export function groupAlerts(objects) {
+  // Step 1: Filter out objects with alertstate 'pending'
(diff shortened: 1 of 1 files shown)


web/src/components/Incidents/processAlerts.js
```diff
@@ -27,10 +27,13 @@
  * // Returns an array where the two alerts are grouped together with deduplicated values.
  */
 export function groupAlerts(objects) {
+  // Step 1: Filter out objects with alertstate 'pending'
+  const filteredObjects = objects.filter((obj) => obj.metric.alertstate !== 'pending');
```
falox, line 31: better `=== 'firing'`

falox (changes requested):

web/src/components/Incidents/processAlerts.js
Fewwy: Done

PR #281, opened 2024-12-04
NO-JIRA: multiple components in tooltip for incidents

Fewwy:


zhuje:
@Fewwy Do you have an JIRA ticket associated with this issue? 
If not title it `NO JIRA: multiple components in tooltip for incidents` so git bot can add the jira/valid-reference label.

PR #288, opened 2024-12-09
OCPBUGS-45896: Fix admin console alert detail graph

PeterYurkovich:
When retrieving the prometheus URL to query, the embedded query-browser was using the passed in namespace in the admin console (which can exist) when it should instead not be passing in any namespace, as the query will go against the whole cluster not a specific namespace

### DIFF
--- a/web/src/components/query-browser.tsx
+++ b/web/src/components/query-browser.tsx
@@ -780,7 +780,7 @@ const QueryBrowser_: React.FC<QueryBrowserProps> = ({
                 {
                   endpoint: PrometheusEndpoint.QUERY_RANGE,
                   endTime: timeRange.endTime,
-                  namespace: perspective === 'dev' ? activeNamespace : namespace,
+                  namespace: perspective === 'dev' ? activeNamespace : '',
                   query,
                   samples: Math.ceil(samples / timeRanges.length),
                   timeout: '60s',


web/src/components/query-browser.tsx
```diff
@@ -780,7 +780,7 @@ const QueryBrowser_: React.FC<QueryBrowserProps> = ({
                 {
                   endpoint: PrometheusEndpoint.QUERY_RANGE,
                   endTime: timeRange.endTime,
-                  namespace: perspective === 'dev' ? activeNamespace : namespace,
+                  namespace: perspective === 'dev' ? activeNamespace : '',
```
jgbernalp, line 783: Maybe something to check on a follow up PR. But do we need then the `namespace` on this component? It seems is only used to trigger new states on change.

PeterYurkovich: We need to trigger new states on change due to the usage in the dev perspective. If the user has queried for something like `cpu_usage` and they use the namespace selector to change namespace they should see the data from the new namespace.

jgbernalp: sure, but then could use the `activeNamespace` change as trigger?

PeterYurkovich: Yes very likely. I'll test out that change in a follow up PR

PeterYurkovich:
The failing test isn't required, this PR just need a LGTM flag 🤦

web/src/components/query-browser.tsx
PeterYurkovich: Follow up PR here: #296. I tested the changes in the developer perspective and the page is able to show the new data when swapping namespace

juzhao:
/test okd-scos-e2e-aws-ovn
if the job is still failed, can we ignore it, it's not required job

juzhao:
see from bug comment, issue is fixed with PR in 4.19
/label qe-approved

PR #289, opened 2024-12-10
OU-615: Enhance Dashboards Page > Dropdown > PersesDashboards with Polling 

zhuje:
### TODO 
- fetch PerseDashboards only when dropdown is open
- clean up and squash commits

### JIRA 
https://issues.redhat.com/browse/OU-615

### Demo/Screenshots 
https://github.com/user-attachments/assets/3b8bc2f2-4a73-4548-b9a8-ee58721404b0
Figure 1. Demo showing polling (video ~1:15 minutes). Dashboards can be added and deleted without having to be manually refreshed. Polling occurs every 15 seconds to update the dashboard drop-down. 

### Testing 
This PR has only been locally tested using the following instructions:
1. Start the Perses development container with this docker command `docker run --name perses -d -p 127.0.0.1:8080:8080 persesdev/perses`. Find more info here: https://github.com/perses/perses?tab=readme-ov-file#docker-images
2. Run the monitoring-plugin locally (e.g `make start-frontend` and in a separate terminal, `make start-console`, then navigate to http://localhost:9000 . The monitoring-plugin will be proxying to the perses instance at localhost:8080) 
3. On OCP navigate to Observe > Dashboards > Dashboards Dropdown. 
4. Now you can add/delete dashboards in the local perses instance at http://localhost:8080. These addition/deletes should be reflected in the Dashboards Dropdown when it polls every 15 seconds.


web/src/components/dashboards/usePerses.ts
```diff
@@ -50,9 +47,9 @@ const reducer = (state: State, action: Action): State => {
   }
 };
 
-export const usePerses = () => {
-  const dashboardsAbort = useRef<() => void | undefined>();
+const URL_POLL_DEFAULT_DELAY = 15000; // 15 seconds
```
jgbernalp, line 50: I think a better solution would be to re fetch the dashboards only when opening the dropdown. Polling consumes additional resources, and in this case the list of dashboards is not constantly available but only when the dropdown opens

PeterYurkovich:
@zhuje you can hold off on further for work this. I don't believe this will be necessary with the direction I am working in. Sorry for not letting you know earlier

PR #290, opened 2024-12-10
NO-JIRA: Run webpack with a different port

upalatucci:
We use some scripts to run monitoring plugin in a different port sometimes. 
It would be cool to have this env in the configuration so we don't have to manually edit the code locally.


jgbernalp (approved):

PeterYurkovich:
/retitle NO-JIRA: Run webpack with a different port

PR #302, opened 2024-12-16
CNV-51386: observe page blank on virt perspective

upalatucci:
We added a new perspective called 'Virtualization'.
In this perspective we added the observe links in the sidebar menu to go directly in the alerts metrics, dashboard and targets page. 
But when we are in this perspective the page is blank. This will fix the issue. 
I'm not sure if you guys want to fix that even for other custom perspective. This is the fix to just make it work for our perspective. 

Steps: 

Screenshot 2024-12-16 at 15 27 26

Screenshot 2024-12-16 at 15 27 30

Screenshot 2024-12-16 at 15 20 47

### DIFF
--- a/web/console-extensions.json
+++ b/web/console-extensions.json
@@ -38,6 +38,35 @@
     }
   },
   {
+    "flags": {
+      "required": ["PROMETHEUS", "MONITORING", "CAN_GET_NS"]
+    },
+    "properties": {
+      "dataAttributes": {
+        "data-quickstart-id": "qs-nav-monitoring"
+      },
+      "id": "observe-virt-perspective",
+      "insertBefore": ["compute-virt-perspective", "usermanagement-virt-perspective"],
+      "name": "%console-app~Observe%",
+      "perspective": "virtualization-perspective"
+    },
+    "type": "console.navigation/section"
+  },
+  {
+    "type": "console.navigation/href",
+    "flags": {
+      "required": ["PROMETHEUS", "MONITORING", "CAN_GET_NS"]
+    },
+    "properties": {
+      "id": "alerting-virt",
+      "name": "%plugin__monitoring-plugin~Alerting%",
+      "href": "/virt-monitoring/alerts",
+      "perspective": "virtualization-perspective",
+      "section": "observe-virt-perspective",
+      "startsWith": ["virt-monitoring/alertrules", "virt-monitoring/silences"]
+    }
+  },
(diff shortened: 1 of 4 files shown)


PeterYurkovich:
@upalatucci Thank you for looking to fix this behavior for your perspective. As we have custom behavior based on the perspective, I don't believe we will want to fix this to work with any possible perspective, as it should be a direct choice of the perspective for them to select which behavior they need. 

This change will also not be enough to completely fix the issue, as the [console-extension.json](https://github.com/openshift/monitoring-plugin/blob/main/web/console-extensions.json#L63) for the monitoring-plugin calls out the admin perspective directly. Although the perspective may not directly switch, the perspectives and extensions would then be out of sync and could lead to other odd or hard to debug issues. Although our code routes most of our perspective code through the usePerspective functions, anything in the openshift-sdk will just use the useActivePerspective which may end up being desynced from the change you are making here. 

I believe that it might be best to look to create new urls and links that are virtualization perspective exclusive, even if they are only showing the admin perspective components. This change would likely need adjustments to the above mentioned console-extension.json file as well as the adjusting the various functions in the usePerspective file, and finally updating the routing components in the alerting.tsx file [here](https://github.com/openshift/monitoring-plugin/blob/main/web/src/components/alerting.tsx#L687).

If you have any questions on how to proceed with all of that feel free to reach out.

web/console-extensions.json
```diff
@@ -32,7 +32,6 @@
       "id": "alerting",
       "name": "%plugin__monitoring-plugin~Alerting%",
       "href": "/monitoring/alerts",
-      "perspective": "admin",
```
PeterYurkovich, line 35: The change needed here isn't the removal of the perspective but rather the addition of virtualization perspective specific links. While the perspective is optional, when it isn't present, it is presumed to be the default perspective ([link](https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/docs/console-extensions.md#consolenavigationhref)) which is the admin perspective.

The link `/monitoring/alerts` should refer to a location in a single perspective. For the developer perspective there is `/dev-monitoring/ns/:ns/alerts`, and for the virtualization perspective there should be a third `/virt-monitoring/alerts` URL.

These new links should then be included in the various usePerspective functions so that all of the pages link together correctly.

PeterYurkovich (changes requested):

PR #304, opened 2024-12-19
OU-433: Split the Dashboard Page between perses and legacy

PeterYurkovich:


PeterYurkovich:
Rebased ontop of #308

PeterYurkovich:
Rebase on top of changes from #313 once the changes have merged to main

web/src/components/console/models/index.ts
```diff
@@ -60,8 +62,33 @@ export const NodeModel = {
   kind: 'Node',
 };
 
-export const NamespaceModel = {
+export const NamespaceModel: K8sModel = {
+  apiVersion: 'v1',
+  label: 'Namespace',
+  // t('public~Namespace')
```
jgbernalp, line 68:
```suggestion
  // t('Namespace')
```

web/src/components/dashboards/dashboard-page.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { usePerspective } from '../hooks/usePerspective';
+
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
+      retry: false,
+      // TODO: Consider if it's possible to link this the refresh interval, which is down 2 layers
+      // in the timespan-dropdown. Or maybe its already in the store
+      refetchInterval: 5 * 1000,
```
jgbernalp, line 26: This might be managed already by the selected option on the dropdown and not in the queryClient

jgbernalp:
@PeterYurkovich looks good to me overall. I noticed the dropdown duplicates the entries by removing spaces, was't able to track it on the code. See an example:

Screenshot 2025-01-14 at 09 51 34

web/src/components/console/models/index.ts
PeterYurkovich: Removed in most recent commit

web/src/components/dashboards/dashboard-page.tsx
PeterYurkovich: The query client only provides the default behavior. I have update the actual usage of the query client to pull the poll interval from the dashboards store.

PeterYurkovich:
The duplication of projects was due to improper merging of the display name and the actual name. These have been fixed in the latest commits.

PeterYurkovich:
Bug found on PR [here](https://issues.redhat.com/browse/OCPBUGS-48472). It has been addressed in the most recent commit.

juzhao:
tested with PR, ([issue](https://issues.redhat.com/browse/OCPBUGS-48472)) is fixed

web/src/components/proxied-fetch.ts
```diff
@@ -65,14 +58,14 @@ export const cancellableFetch = <T>(
...
-    return { request: () => fetchPromise, abort };
+    return fetchPromise;
   }
 
   const timeoutPromise = new Promise<T>((_resolve, reject) => {
     setTimeout(() => reject(new TimeoutError(url.toString(), timeout)), timeout);
   });
 
-  const request = () => Promise.race([fetchPromise, timeoutPromise]);
+  const request = Promise.race([fetchPromise, timeoutPromise]);
```
jgbernalp, line 68: We might need to keep this as it is as the thunk allows to start the timeout at the moment the request is executed rather than when the proxied fetch is called

PR #306, opened 2024-12-20
Split alertrulesdetailpage out into separate file

PeterYurkovich:
/hold

Since [OU-561](https://issues.redhat.com/browse/OU-561) was completed without finishing the refactor, this PR looks to finalize the Alerting refactor. Pulling out the final two pages into their own files, and removing the routing entrypoint of the monitoring UI from the alerting.tsx file to a new router.tsx file.

We should hold this PR until everyone is back in January

PR #307, opened 2025-01-02
OCPBUGS-47535: use pf v5 variables and table component

jgbernalp:
This PR:

- Replaces all usages of patternfly V4 with V5
- Updates the patternfly table from V4 to V5
- Upgrades patternfly react table and charts packages

Smoke tested with https://github.com/openshift/console/pull/14615


PeterYurkovich:
@jgbernalp This PR is going to run into some conflicts with both #306 and #304. We should decide which order we want to merge the items in (I believe this PR can come first) and then rebase the others PR's on top of that one. 

Not really relevant, but this situation would be a lovely place for using something like [PR stacking](https://www.stacking.dev/)

PR #308, opened 2025-01-03
NO-JIRA: Account for store version drift

PeterYurkovich:
As the `openshift/console` currently shackles the `monitoring-plugin` store, so too will the `monitoring-plugin` store shackle the `monitoring-console-plugin` store, as the `monitoring-console-plugin` will deployed across OCP versions while the `monitoring-plugin` version will be pegged to the OCP version. This change looks to plan for the future so separate out the `monitoring-console-plugin` store to a unique name, so that when the store is removed from the `openshift/console` and brought into the `monitoring-plugin` they will have unique but understandable names.

I'm also open to changing the `monitoring-console-plugin` usage in the PR to instead be `mcp` so that we don't have the (kinda ugly) square bracket syntax

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Once the code has been updated, make sure to update the helm chart and variables
 
 ### Redux Store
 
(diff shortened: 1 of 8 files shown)


PeterYurkovich:
The changes here only impact the acm-alerting, perses-dashboards, and incidents features. Do we want to try to get this change in before the COO 1.0.0 release for acm-alerting? I think we are basically in feature freeze for updating the digest in the [rhobs/konflux-coo](https://github.com/rhobs/konflux-coo) repo, so I don't think it should be a big deal if it only makes it in for 1.1.0.

We can wait for regression testing, but there shouldn't be a code path that is accessible for this change in any versions of our product offerings

PR #314, opened 2025-01-08
OCPBUGS-48142: Fix Dev Perspective Alert URL namespace

PeterYurkovich:
Fixes an issue which prevented the namespace from being populated in the alert URL located in the silences page.

This was due to the namespace not being passed in to the URL generator function


juzhao:
tested with PR, issue is fixed, see bug comment
/label qe-approved

PR #316, opened 2025-01-09
RHINENG-15190: fix gap in incident charts

Fewwy:
updated logic for organizing intervals of the same severity for the bar chart generation

### DIFF
--- a/web/src/components/Incidents/utils.js
+++ b/web/src/components/Incidents/utils.js
@@ -83,9 +83,36 @@ export const createIncidentsChartBars = (incident) => {
   return data;
 };
 
-export const createAlertsChartBars = (alert) => {
-  const groupedData = consolidateAndMergeIntervals(alert);
+function groupAlertTimestamps(data) {
+  if (data.length === 0) return [];
 
+  let result = [];
+  let start_time = data[0][0];
+  let current_value = data[0][1];
+
+  for (let i = 1; i < data.length; i++) {
+    let timestamp = data[i][0];
+    let value = data[i][1];
+
+    // If the second index value changes
+    if (value !== current_value) {
+      // Push the grouped data into the result
+      result.push([start_time, data[i - 1][0], current_value]);
+
+      // Start a new group
+      start_time = timestamp;
+      current_value = value;
+    }
+  }
+
+  // Push the last group
+  result.push([start_time, data[data.length - 1][0], current_value]);
+
+  return result;
+}
+
+export const createAlertsChartBars = (alert) => {
+  const groupedData = groupAlertTimestamps(alert.values);
   const data = [];
   for (let i = 0; i < groupedData.length; i++) {
     data.push({
--- a/web/src/components/Incidents/utils.js
+++ b/web/src/components/Incidents/utils.js
@@ -7,7 +7,7 @@ import { setIncidentsActiveFilters } from '../../actions/observe';
 function consolidateAndMergeIntervals(data) {
   const severityRank = { 2: 2, 1: 1, 0: 0 };
 
-  //Eliminate overlapping timestamps with lower severities
+  // Eliminate overlapping timestamps with lower severities
   const highestSeverityValues = data.values.reduce((acc, [timestamp, severity]) => {
     if (!acc[timestamp] || severityRank[severity] > severityRank[acc[timestamp]]) {
       acc[timestamp] = severity;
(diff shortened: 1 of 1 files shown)


zhuje:
I think there is a logic mistake here, correct me if I'm wrong... In `getSeverityName()` this be `value === 2` instead of `value === '2'`, etc.? Because in `consolidateAndMergeIntervals()` you set the value as a number instead of string ` const severityRank = { 2: 2, 1: 1, 0: 0 }`. 
``` 
  const getSeverityName = (value) => {
    return value === '2' ? 'Critical' : value === '1' ? 'Warning' : 'Info';
  };

Suggested Change: 
  const getSeverityName = (value) => {
    return value === 2 ? 'Critical' : value === 1 ? 'Warning' : 'Info';
  };
```


Also, a nit you have a there are lot of if/else statements in `createIncidentschartsBars()` such as `getSeverityName` and the `data.[x].fill` property. Do you think using switch statements here would make sense?  I have an example of the changes below (these haven't been tested). 

```
const createIncidentsChartBars = (incident) => {
  const groupedData = consolidateAndMergeIntervals(incident);
  console.log({groupedData})

  const data = [];
//   const getSeverityName = (value) => {
//     return value === '2' ? 'Critical' : value === '1' ? 'Warning' : 'Info';
//   };

    const getSeverityProps = (severityValue) => {
    switch (severityValue) {
        case 2:
            return {name: 'Critical', fill: global_danger_color_100.var}
        case 1:
            return {name: 'Warning', fill: global_warning_color_100.var};
        case 0:
            return {name: 'Info', fill: global_info_color_100.var}
        default:
            return {name: 'Undefined', fill: global_info_color_100.var};
    }
    };

  for (let i = 0; i < groupedData.length; i++) {
    const severityValue = groupedData[i][2]
    const severity = getSeverityProps(severityValue);

    console.log({severity})

    data.push({
      y0: new Date(groupedData[i][0]),
      y: new Date(groupedData[i][1]),
      x: incident.x,
      name: severity.name,
      componentList: incident.componentList || [],
      group_id: incident.group_id,
      fill: severity.fill
    });
  }

  return data;
};
```

Fewwy:
> I think there is a type mistake here; correct me if I'm wrong... In `getSeverityName()`, should this be `value === 2` instead of `value === '2'`, etc.? Because in `consolidateAndMergeIntervals()`, you set the value as a number instead of string ` const severityRank = { 2: 2, 1: 1, 0: 0 }`.
> 
> ```
>   const getSeverityName = (value) => {
>     return value === '2' ? 'Critical' : value === '1' ? 'Warning' : 'Info';
>   };
> 
> Suggested Change: 
>   const getSeverityName = (value) => {
>     return value === 2 ? 'Critical' : value === 1 ? 'Warning' : 'Info';
>   };
> ```

@zhuje Check the API response for the timestamps. It looks like this  [1733778756.253,"1"]

> Also, a nit you have a lot of if/else statements in `createIncidentschartsBars()` such as `getSeverityName` and the `data.[x].fill` property. Would using switch statements here make sense? I have an example of the changes below (these haven't been tested).
> 
> ```
> const createIncidentsChartBars = (incident) => {
>   const groupedData = consolidateAndMergeIntervals(incident);
>   console.log({groupedData})
> 
>   const data = [];
> //   const getSeverityName = (value) => {
> //     return value === '2' ? 'Critical' : value === '1' ? 'Warning' : 'Info';
> //   };
> 
>     const getSeverityProps = (severityValue) => {
>     switch (severityValue) {
>         case 2:
>             return {name: 'Critical', fill: global_danger_color_100.var}
>         case 1:
>             return {name: 'Warning', fill: global_warning_color_100.var};
>         case 0:
>             return {name: 'Info', fill: global_info_color_100.var}
>         default:
>             return {name: 'Undefined', fill: global_info_color_100.var};
>     }
>     };
>   for (let i = 0; i < groupedData.length; i++) {
>     const severityValue = groupedData[i][2]
>     const severity = getSeverityProps(severityValue);
> 
>     console.log({severity})
> 
>     data.push({
>       y0: new Date(groupedData[i][0]),
>       y: new Date(groupedData[i][1]),
>       x: incident.x,
>       name: severity.name,
>       componentList: incident.componentList || [],
>       group_id: incident.group_id,
>       fill: severity.fill
>     });
>   }
> 
>   return data;
> };
> ```

This part will be rewritten soon because I need to adjust fill values depending on the Dark theme or Light theme for openshift.

zhuje:
>  Check the API response for the timestamps. It looks like this [1733778756.253,"1"]

Ah, that makes sense. 
Then nit, should we clarify `const severityRank = { "2": 2, "1": 1, "0": 0 };`

I looked it up and JavaScript automatically coerces the string "1" to the number 1 when accessing the object.

So, this technically it still works as is  with `const severityRank = { 2: 2, 1: 1, 0: 0 };`. 

> This part will be rewritten soon because I need to adjust fill values depending on the Dark theme or Light theme for openshift.

Fine to leave it as is if its going to be rewritten soon.

Fewwy:
@zhuje sorry but I want to focus on more major things for now :)

zhuje:
> @zhuje sorry but I want to focus on more major things for now :)

That's fine. Its a small nit. 

/lgtm

zhuje:
The code looks good to me. After @sthundat qe-approves, I'll mark it as lgtm. 

@sthundat can you please test?

PR #317, opened 2025-01-09
OU-630: Add Needed Virtualization Perspective Changes

PeterYurkovich:
This change adds a new redux reducer which is exclusive to the monitoring-plugin, as well as the extension points needed for the virtualization perspective.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Once the code has been updated, make sure to update the helm chart and variables
 
 ### Redux Store
 
-Since the store for the `monitoring-plugin` is stored in the `openshift/console` codebase and updates to the store that are aren't tied directly to the OCP are needed, when the default extension points are removed due to the presense of a feature flag a duplicate store is created at the `.state.plugins.mcp` path. A combination of the `useFeatures` hook and the `getLegacyObserveState` (which is dependant on the perspective) can be used to retrieve the state from the redux store based on the mode the plugin was deployed in.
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -128,7 +128,7 @@ Once the code has been updated, make sure to update the helm chart and variables
 
 ### Redux Store
 
-Since the store for the `monitoring-plugin` is stored in the `openshift/console` codebase and updates to the store that are aren't tied directly to the OCP are needed, when the default extension points are removed due to the presense of a feature flag a duplicate store is created at the `.state.plugins.monitoring` path. A combination of the `useFeatures` hook and the `getObserveState` (which is dependant on the perspective) can be used to retrieve the state from the redux store based on the mode the plugin was deployed in.
+Since the store for the `monitoring-plugin` is stored in the `openshift/console` codebase and updates to the store that are aren't tied directly to the OCP are needed, when the default extension points are removed due to the presense of a feature flag a duplicate store is created at the `.state.plugins.mcp` path. A combination of the `useFeatures` hook and the `getLegacyObserveState` (which is dependant on the perspective) can be used to retrieve the state from the redux store based on the mode the plugin was deployed in.
```
juzhao, line 131:
```suggestion
Since the store for the `monitoring-plugin` is stored in the `openshift/console` codebase and updates to the store that are aren't tied directly to the OCP are needed, when the default extension points are removed due to the presence of a feature flag a duplicate store is created at the `.state.plugins.mcp` path. A combination of the `useFeatures` hook and the `getLegacyObserveState` (which is dependent on the perspective) can be used to retrieve the state from the redux store based on the mode the plugin was deployed in.
```

juzhao:
tested with PR, no issue, see bug comment

PR #319, opened 2025-01-15
RHINENG-15284: Sync alerts chart and incidents table with days filter

Fewwy:
The issue https://issues.redhat.com/browse/OU-628

This PR fixes 2 major things:
1) I updated the logic for creating bars for the alert chart. Now the bar respects the gaps in time and the bars are created accordingly and not smushed all together into 1 line
2) I added more logic to how we manage loading states for the alerts chart and incidents table. Now if the date changes -> we force loading states so the component won't render BEFORE we get new data from the API. Previously there was an issue that was causing the component to render again with old data, without waiting for the API calls to resolve.

Smaller things I added in this PR:
I added unselect logic to the "onclick" function that fires when the user clicks on an incident in the incident chart.
I replaced the PF color variables because of the bug when PF loses its styles and these variables doesn't work

### DIFF
--- a/web/src/components/Incidents/AlertsChart/AlertsChart.jsx
+++ b/web/src/components/Incidents/AlertsChart/AlertsChart.jsx
@@ -1,4 +1,3 @@
-/* eslint-disable prettier/prettier */
 import * as React from 'react';
 
 import { Chart, ChartAxis, ChartBar, ChartGroup, createContainer } from '@patternfly/react-charts';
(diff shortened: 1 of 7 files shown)


web/src/components/Incidents/AlertsChart/AlertsChart.jsx
```diff
@@ -1,26 +1,33 @@
+/* eslint-disable prettier/prettier */
```
jgbernalp, line 1: We should not disable these rules as they maintain our code styling.

web/src/components/Incidents/AlertsChart/AlertsChart.jsx
```diff
@@ -74,9 +81,9 @@ const AlertsChart = ({ alertsData = [], chartDays }) => {
               }
               domainPadding={{ x: [30, 25] }}
               legendData={[
-                { name: 'Critical', symbol: { fill: global_danger_color_100.var } },
-                { name: 'Info', symbol: { fill: global_info_color_100.var } },
-                { name: 'Warning', symbol: { fill: global_warning_color_100.var } },
+                { name: 'Critical', symbol: { fill: '#c9190b' } },
```
jgbernalp, line 84: Are these colors compatible with dark mode?

falox: I think colors will change anyway with https://issues.redhat.com/browse/OU-634 (see comments)

jgbernalp: Then at least make sense to define this as constants and share them to avoid mistyping or using a different color in different places.

web/src/components/Incidents/AlertsChart/AlertsChart.jsx
jgbernalp: @Fewwy Please fix this as this rules should not be disabled.

web/src/components/Incidents/AlertsChart/AlertsChart.jsx
jgbernalp: @Fewwy ok, please fix the issue above. Maybe in the future we can try to include changes for unrelated tasks in a specific PR so they don't block bigger changes.

PR #322, opened 2025-01-16
RHINENG-15362: Sort the bars in the charts based on their first timestamp

Fewwy:
https://issues.redhat.com/browse/RHINENG-15362
Sort the bars in the charts based on their first timestamp

### DIFF
--- a/web/src/components/Incidents/IncidentsPage.jsx
+++ b/web/src/components/Incidents/IncidentsPage.jsx
@@ -315,7 +315,7 @@ const IncidentsPage = () => {
                   }
                 />
               </ToolbarItem>
-              <ToolbarItem className="pf-v5-m-align-right">
+              <ToolbarItem className="pf-m-align-right">
                 <Button
                   variant="link"
                   icon={hideCharts ? <CompressArrowsAltIcon /> : <CompressIcon />}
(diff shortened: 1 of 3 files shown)


zhuje:
looks good to me, look label once QE tests and approves. 

@sthundat Could you test this, please?

web/src/components/Incidents/IncidentsPage.jsx
```diff
@@ -315,7 +315,7 @@ const IncidentsPage = () => {
                   }
                 />
               </ToolbarItem>
-              <ToolbarItem className="pf-v5-m-align-right">
+              <ToolbarItem className="pf-m-align-right">
```
jgbernalp, line 318: The current version supported for the plugin is PF5. Is this a custom class or needs patternfly css?

Fewwy: If you use className="pf-v5-m-align-right" this is the result -> it doesn't work
image

There is no classname like this in pf5, it's a mistake I had to fix to make the alignment work again.

jgbernalp: Then maybe we should use the component props instead of applying css: https://v5-archive.patternfly.org/components/toolbar#toolbaritem

Fewwy: @jgbernalp I updated css with prop for toolbar item

jgbernalp (approved):

PR #345, opened 2025-02-10
upgrade go version with security fixes

jgbernalp:


zhuje:
Maybe we need to `go mod tidy` to sync `go.mod` and `go.sum`?

PeterYurkovich:
Max go version is dependent on the base image used in the various Dockerfiles. The max go version on the `quay.io/redhat-cne/openshift-origin-release:rhel-9-golang-1.22-openshift-4.17 as go-builder` is `1.22.1`. We will need to update the base images in addition to the changes in `go.mod`

jgbernalp:
The base image needs to be updated first. Then we can include this change to be explicit regarding the go version for the base images when building the plugin. For now we need to wait until the base image is patched.

PR #347, opened 2025-02-13
OU-643: fix filter dropdpdowns to match patterfly correct use, remove deprecated dropdown components

jgbernalp:


falox:
I tested, the situation has defilitely improved but there is still something to fix:

- the days filter does not react to clicks
- (minor) the first time you click on the first item of the filters, the labels don't respond to clicks (you have to click on the checkbox); the other items respond to clicks on both the checkbox and the label. After you close and reopen the drop-down, the first item works correctly.

falox:
@jgbernalp I tested the last commit and now it works!

PR #349, opened 2025-02-17
rhineng-16008: add cutoff for alerts timestamps

Fewwy:
Due to the limitations of how the API works when we request alerts data we can "catch" alerts data for the other incidents. Specifically the time before/after the chosen/selected incident starts/ends. To avoid showing this data I need to introduce a cut-off time, and make sure we're not showing alerts for the time when the chosen incident is not "happening"

Before
image

After
image


PeterYurkovich:
/retitle rhineng-16008: add cutoff for alerts timestamps

PR #353, opened 2025-02-24
OU-618: Various Perses Updates

PeterYurkovich:
# Second set of changes for Perses

These various commits fix issues found in Evelyn's testing [doc](https://docs.google.com/document/d/1Rdy47M2VWk9O53Y_rnj1WQ9UIGgsBTrliDm1q-ntiKQ/edit?tab=t.0#heading=h.3q6mi1fk871), remove the redux dependency, address the flickering issue on refetch (kinda) and standardizes URL parameter adjustments on the `use-query-param` pacakge.

First commit has a primary focus on removing the redux dependency for perses dashboards and fixing the dashboards url to be a v2 rather than `perses-dashboards`. It also changes the URL from having the format from `dashboards/v2/{dashboard_name}` to `dashboards/v2?dashboard={dashboard_name}`. Since there are now project as containers for lists of dashboards it doesn't make sense to have the `{dashboard_name}` as the root parameter. 

In the future we can discuss if `dashboards/v2/{project_name}/{dashboard_name}` will be better, but that will depend on the UX flows decided upon for the customizable dashboards. 

---

Second commit removes a duplicate poller for alerts. As soon as we can it would be best to just centralize on react-query to not have to manually handle all the polling.

---

Third commit fixes a bug which prevented variables from loading when refreshing the legacy dashboards pages. The basic flow of the bug was that the when selecting `Observe > Dashboards` or when selecting another dashboard the `changeLegacyDashboard` function is called. But because the dashboardName is already selected when refreshing, the function isn't called. A first load variable has been created to call the function on the first load.

---

Fourth commit adds virt perspective extension points and changes _Perses Dashboards_ to _Dashboards (Perses)_ in perspective where both dashboard pages are available (admin, virt) but leaves the title as _Dashboards_ in the acm perspective.

---

Fifth commit fixes an issue which didn't take the dashbaord url parameter into account when using the developer perspective.

---

Sixth commit fixes a typo and updates the "dashboard exists but is empty" state so that it doesn't mention an edit button.

---

Seventh commit fixes an issue with reactivity caused by setting query parameters outside of `use-query-params`. Although the new parameters would be found on a refresh, their changes weren't recognized for reactivity. This led to odd behavior such as the new duration of a time range being reactive, but the "deleteEndTime" not causing the end time to also update.

---

Eighth commit sets the default refresh time to "No refresh" after inspecting perses, the issue with flickering when refreshing comes from upstream. We should look to solve this issue upstream in the future, but set the refresh interval to default to off for now to prevent flickering from happening frequently.


config/perses-dashboards.patch.json
```diff
@@ -50,12 +50,42 @@
...
+    }
+  },
+  {
+    "op": "add",
+    "path": "/extensions/1",
+    "value": {
+      "type": "console.navigation/href",
+      "flags": {
+        "required": ["PROMETHEUS", "MONITORING", "CAN_GET_NS"]
+      },
```
jgbernalp, line 80: These flags might not be required anymore as regular users could also access the dashboards they are allowed to.

web/src/components/dashboards/legacy/legacy-dashboard-page.tsx
```diff
@@ -64,10 +65,12 @@ const MonitoringLegacyDashboardsPageWrapper: React.FC<MonitoringLegacyDashboards
   match,
 }) => {
   return (
-    <MonitoringLegacyDashboardsPage_
-      urlBoard={match.params.dashboardName}
-      namespace={match.params?.ns}
-    />
+    <PersesContext.Provider value={false}>
```
jgbernalp, line 68: Why do we need a Perses context in legacy dashboards?

web/src/components/alerting/AlertsDetailPage.tsx
```diff
@@ -128,20 +120,6 @@ const AlertsDetailsPage_: React.FC<AlertsDetailsPageProps> = ({ history, match }
 
   const AlertsChart = alertsChart?.[0];
 
-  const [customExtensions] =
```
jgbernalp, line 131: Was this custom extension alert source, moved elsewhere. This allows log based alerts to have a detail page.

config/perses-dashboards.patch.json
PeterYurkovich: While true, this first round doesn't have RBAC applied to the individual projects and dadhboards so I think we should leave the RBAC on the dashboards menu

web/src/components/dashboards/legacy/legacy-dashboard-page.tsx
PeterYurkovich: The context provides a "isPerses" boolean to deeply nested components. Specifically the time dropdowns. Changing their behavior from using the redux store to using the URL params

PR #356, opened 2025-02-28
OU-499: remove deprecated components

jgbernalp:
This PR removes deprecated components and replaces them with the current versions


jgbernalp:
/label ux-approved
No UX changes

PR #359, opened 2025-03-11
incident utils typescript conversion

Fewwy:
Converted JS to TS with more typings and small updates of some of the functions

Removed some of the functionality for Days filters that work with filter chips, it was unnecessary because days filter doesn't create filter chips.

### DIFF
--- a/web/src/components/Incidents/utils.ts
+++ b/web/src/components/Incidents/utils.ts
@@ -5,13 +5,63 @@ import global_danger_color_100 from '@patternfly/react-tokens/dist/esm/global_da
 import global_info_color_100 from '@patternfly/react-tokens/dist/esm/global_info_color_100';
 import global_warning_color_100 from '@patternfly/react-tokens/dist/esm/global_warning_color_100';
 
+type Timestamps = [Date, string];
+
+type SpanDates = [Date];
+
+type Theme = 'dark' | 'light';
+
+type AlertsIntervalsArray = [Date, Date, 'data' | 'nodata'];
+
+type Incident = {
+  component: string;
+  componentList: Array<string>;
+  critical: boolean;
+  informative: boolean;
+  warning: boolean;
+  resolved: boolean;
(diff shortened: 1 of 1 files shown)


web/src/components/Incidents/utils.ts
```diff
@@ -168,22 +230,24 @@ export const createIncidentsChartBars = (incident, theme, dateArray) => {
...
 
-  const intervals = [];
+  const intervals: Array<AlertsIntervalsArray> = [];
   let currentStart = sortedValues[0][0],
     previousTimestamp = new Date(currentStart);
 
   for (let i = 1; i < sortedValues.length; i++) {
     const currentTimestamp = new Date(sortedValues[i][0]);
-    const timeDifference = (currentTimestamp - previousTimestamp) / 60000; // Convert to minutes
+    const timeDifference = (currentTimestamp.valueOf() - previousTimestamp.valueOf()) / 60000; // Convert to minutes
```
jgbernalp, line 244: We should probably use one of `valueOf()` or `getTime()` but not both since the function returns the same value. I'm in favor of sticking with `getTime()` as is more clear what the intention is.

web/src/components/Incidents/utils.ts
```diff
@@ -393,29 +445,22 @@ export const onDeleteIncidentFilterChip = (type, id, filters, setFilters) => {
...
-      setIncidentsActiveFilters({
-        incidentsActiveFilters: {
-          incidentFilters: filters.incidentFilters,
-          days: [],
-        },
-      }),
-    );
-  }
+export const onDeleteGroupIncidentFilterChip = (
+  type: string,
```
jgbernalp, line 449: it seems type is not used here anymore

web/src/components/Incidents/utils.ts
Fewwy: Thanks I will update the code

PR #363, opened 2025-03-17
COO-759: update datasource whenerver the csrf token changes

PeterYurkovich:
This PR looks to fix an issue which prevented the datasources from updating whenever a CSRF cookie was changed, thus leading to authentication errors.


PeterYurkovich:
No UX changes

jgbernalp:
@PeterYurkovich While testing this, I found an issue here: https://github.com/openshift/monitoring-plugin/blob/main/web/src/components/dashboards/perses/datasource-api.ts#L53 This list can return successfully but empty, mind checking?

PeterYurkovich:
@jgbernalp I threw a new error (which is caught by the perses error boundry) if there wasn't a matching datasource. I also updated some console based translation namespaces and some items which were missing t() calls and then reran the npm run i18n command to update the json file

New image shows the more descriptive error message

image

PR #365, opened 2025-03-18
COO-768: add plugin proxy to perses proxy path

jgbernalp:
This PR, adds the plugin proxy path before the perses proxy so requests to datasources are proxied correctly

### DIFF
--- a/web/src/components/dashboards/perses/datasource-api.ts
+++ b/web/src/components/dashboards/perses/datasource-api.ts
@@ -5,7 +5,7 @@ import { fetchGlobalDatasourceList } from './perses/global-datasource-client';
 import { TFunction } from 'i18next';
 
 export class OcpDatasourceApi implements DatasourceApi {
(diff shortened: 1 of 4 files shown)


web/src/components/dashboards/perses/datasource-api.ts
```diff
@@ -5,7 +5,9 @@ import { fetchGlobalDatasourceList } from './perses/global-datasource-client';
 import { TFunction } from 'i18next';
 
 export class OcpDatasourceApi implements DatasourceApi {
-  constructor(public csrfToken: string, public t: TFunction) {}
+  constructor(public csrfToken: string, public t: TFunction, public basePath: string) {
+    this.basePath = basePath;
```
PeterYurkovich, line 9: Setting the parameter as public means we don't have to set it in the constructor

PeterYurkovich (changes requested):

PR #367, opened 2025-03-20
OU-700: don't show metrics links in acm perspective

PeterYurkovich:
Don't show metrics links in the acm perspective since the metrics page isn't present


PeterYurkovich:
Only UX change is removing links that lead no-where

PR #369, opened 2025-03-25
OCPBUGS-52592: Fix "Export as CSV"

zhuje:
JIRA 
https://issues.redhat.com/browse/OCPBUGS-52592

### Context 
`const isSpan = (item) => item?.title?.props?.children;`
The optional chaining operator `?.` gracefully handles undefined `item` by not throwing errors and allowing `item?.title?.props?.children` to result as `undefined`.

### DIFF
--- a/web/src/components/metrics.tsx
+++ b/web/src/components/metrics.tsx
@@ -459,7 +459,7 @@ const QueryKebab: React.FC<{ index: number }> = ({ index }) => {
     dispatch(queryBrowserDuplicateQuery(index));
   }, [dispatch, index]);
 
-  const isSpan = (item) => item.title?.props?.children;
+  const isSpan = (item) => item?.title?.props?.children;
   const getSpanText = (item) => item.title.props.children;
 
   // Takes data from QueryTable and removes/replaces all html objects from columns and rows


web/src/components/metrics.tsx
```diff
@@ -459,8 +459,8 @@ const QueryKebab: React.FC<{ index: number }> = ({ index }) => {
     dispatch(queryBrowserDuplicateQuery(index));
   }, [dispatch, index]);
 
-  const isSpan = (item) => item.title?.props?.children;
-  const getSpanText = (item) => item.title.props.children;
+  const isSpan = (item) => item?.title?.props?.children;
+  const getSpanText = (item) => item?.title?.props?.children;
```
PeterYurkovich, line 463: Doesn't need to be optionally chained since it is only called if children exists in the isSpan function

PeterYurkovich (changes requested):

web/src/components/metrics.tsx
zhuje: You're right, removed in the latest push.

PeterYurkovich:
no UX changes

PeterYurkovich (approved):

web/src/components/metrics.tsx
```diff
@@ -459,7 +459,7 @@ const QueryKebab: React.FC<{ index: number }> = ({ index }) => {
     dispatch(queryBrowserDuplicateQuery(index));
   }, [dispatch, index]);
 
-  const isSpan = (item) => item.title?.props?.children;
+  const isSpan = (item) => item?.title?.props?.children;
```
jgbernalp, line 462: To fix in a follow up PR, but we should implement this as a [Type Guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards), this way typescript could have caught the issue. Also we need to define the right type on the getSpanText function so it does not depend on external logic to catch errors.

etmurasaki (approved):

PR #373, opened 2025-03-28
WIP:  OCPBUGS-52592: Add typeguard to metrics export csv

zhuje:
JIRA
https://issues.redhat.com/browse/OCPBUGS-52592

### DIFF
--- a/web/src/components/metrics.tsx
+++ b/web/src/components/metrics.tsx
@@ -459,7 +459,16 @@ const QueryKebab: React.FC<{ index: number }> = ({ index }) => {
     dispatch(queryBrowserDuplicateQuery(index));
   }, [dispatch, index]);
 
-  const isSpan = (item) => item?.title?.props?.children;
+  type Item = {
+    title?: {
+      props?: {
+        children?: any;
+      };
+    };
+  };
+  const isSpan = (item: any): item is Item => {
+    return !!item?.title?.props?.children;
+  };
   const getSpanText = (item) => item.title.props.children;
 
   // Takes data from QueryTable and removes/replaces all html objects from columns and rows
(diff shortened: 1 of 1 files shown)


zhuje:
Enhancement of https://github.com/openshift/monitoring-plugin/pull/369

web/src/components/metrics.tsx
```diff
@@ -459,7 +459,16 @@ const QueryKebab: React.FC<{ index: number }> = ({ index }) => {
...
 
-  const isSpan = (item) => item?.title?.props?.children;
+  type Item = {
+    title?: {
+      props?: {
+        children?: any;
+      };
+    };
+  };
+  const isSpan = (item: any): item is Item => {
```
PeterYurkovich, line 469: I think we should look to type the queryTableData and columnHeader info rather than declaring the type on the function only.

web/src/components/metrics.tsx
```diff
@@ -459,23 +472,30 @@ const QueryKebab: React.FC<{ index: number }> = ({ index }) => {
     dispatch(queryBrowserDuplicateQuery(index));
   }, [dispatch, index]);
 
-  const isSpan = (item) => item?.title?.props?.children;
+  type Item = {
+    title?: {
```
jgbernalp, line 476: we should avoid using any, I think in this case we know that title a ReactElement, which will include all other needed properties.

web/src/components/metrics.tsx
```diff
@@ -723,14 +743,15 @@ export const QueryTable: React.FC<QueryTableProps> = ({ index, namespace, custom
...
-    columns = [
-      '',
-      ...allLabelKeys.map((k) => ({
-        title: <span>{k === '__name__' ? t('Name') : k}</span>,
-        transforms,
-      })),
-      { title: t('Value'), transforms },
-    ];
+    const btnColumnPlaceholder = '';
+    const titleColumn: TitleColumn[] = allLabelKeys.map((k) => ({
```
jgbernalp, line 747:
```suggestion
    const columnsTitle: TitleColumn[] = allLabelKeys.map((k) => ({
```

PR #374, opened 2025-04-01
OCPBUGS-54211: fix issue preventing dev silences from loading

PeterYurkovich:
The alerts were actually being fetched correctly. However the silences state was being pulled from the admin perspective state store, which was failing (correctly) to poll the non-tenancy API for the notification bell. In addition, the silences poller wasn't being called on all pages that needed it.

The tenancy API is not returning alerts, which is addressed in openshift/cluster-monitoring-operator#2585


PeterYurkovich:
Base image has been deleted. I'm working to update all of the base images across our plugins and branches

PeterYurkovich:
/label ux-approved
No UX changes other than fixing a visual bug

zhuje (approved):
/lgtm

PR #375, opened 2025-04-02
OU-724: remove dependency on deleted image

PeterYurkovich:


zhuje (approved):

PeterYurkovich:
/label qe-approved
/label ux-approved

No ux needed for image changes. No qe needed since automation testing will catch issues

PR #379, opened 2025-04-11
OU-678: fix: allow to set refresh interval to off

jgbernalp:
This allows to set the auto refresh to off. Taken from the pf6 changes to be able to backport it.


jgbernalp:
/label ux-approved

no UX changes

PR #392, opened 2025-04-22
OU-729,OU-723: Add label typeahead

PeterYurkovich:
Re-add label typeahead to the filters by only aggregating the alerts after the filtering has been done so that the labels on the individual alerts are available to be filtered on.

https://github.com/user-attachments/assets/fab156e8-5494-4c3b-a6c6-16264a16bed1

This PR also looks to add cluster name truncation after 50 characters to prevent dialog sprawl across the page.
Screenshot From 2025-04-22 15-52-25


PeterYurkovich:
/label ux-approved
fixing bug, no UX review needed

PR #396, opened 2025-04-23
OU-767: Allow filtering by cluster name

PeterYurkovich:
This PR looks to recover from a regression of a previous bug. It adds back a type so that the filter is able to be used, and then also allows for all items to be matched when no filters are set. 

https://github.com/user-attachments/assets/49092fb3-65e7-4b9b-805f-fe52d87fc1fe


PeterYurkovich:
/label ux-approved
/label qe-approved
Bugfix so no UX needed. QE will be completed post merge

PR #398, opened 2025-04-24
NO-JIRA: fix(incidents): performance improvements for incidents page

Fewwy:
1. Add memoization for parse url params function because it was called too often.
2. Wrapped chart data calculations in memoization to avoid recalculations if the values are the same
3. Wrapped getOpacity and isHidden in useCallback to reduce the number of rerenders

### DIFF
--- a/web/src/components/Incidents/AlertsChart/AlertsChart.jsx
+++ b/web/src/components/Incidents/AlertsChart/AlertsChart.jsx
@@ -24,7 +24,6 @@ import {
 
 const AlertsChart = ({ chartDays, theme }) => {
   const dispatch = useDispatch();
-  const [chartData, setChartData] = React.useState([]);
   const [chartContainerHeight, setChartContainerHeight] = React.useState();
   const [chartHeight, setChartHeight] = React.useState();
   const alertsData = useSelector((state) =>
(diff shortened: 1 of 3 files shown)


web/src/components/Incidents/AlertsChart/AlertsChart.jsx
```diff
@@ -39,43 +38,39 @@ const AlertsChart = ({ chartDays, theme }) => {
...
 
   const [width, setWidth] = React.useState(0);
   const containerRef = React.useRef(null);
-  const handleResize = () => {
+  const handleResize = React.useCallback(() => {
     if (containerRef.current && containerRef.current.clientWidth) {
       setWidth(containerRef.current.clientWidth);
     }
-  };
+  }, []);
```
jgbernalp, line 68: Do we need a dependency on the `containerRef.current`? this might start as null and then populated with the reference after the first render.

Fewwy: The .current value of a ref is mutable and doesn't trigger re-renders
if I try to add it, React warns me about it
image

jgbernalp:
/label ux-approved
No UX changes

jgbernalp:
/label qe-approved
QE can happen post merge

PR #402, opened 2025-04-28
OU-774: bound table scroll and column alignment

PeterYurkovich:
This PR looks to fix an issue causing columns to be right aligned. This was triggered by the usage of the `<Th modifier="fitContent">` parameter which was added to fix [OCPBUGS-54292](https://issues.redhat.com/browse/OCPBUGS-54292).

We swap to using the `nowrap` which shows the full column title and prevents the right alignment.

Additionally this PR binds the table to the page size, placing a horizontal scroll bar around only the table when neccessary.

https://github.com/user-attachments/assets/ebe09fce-3339-4fdb-8299-78ce2ee4afd0

### DIFF
--- a/web/src/components/MetricsPage.tsx
+++ b/web/src/components/MetricsPage.tsx
@@ -42,6 +42,7 @@ import {
 } from '@patternfly/react-core';
 import { ChartLineIcon, CompressIcon } from '@patternfly/react-icons';
 import {
+  InnerScrollContainer,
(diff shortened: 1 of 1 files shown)


PeterYurkovich:
/label ux-approved
/label qe-approved
No UX for bugfix PR, QE to be completed post-merge

PeterYurkovich:
/unhold
branching is done

PR #404, opened 2025-04-29
OU-745: Check for CSRF change on each fetch cycle

PeterYurkovich:
This PR looks to manually memoize the the datasource API object so that the CSRF token will be checked for changes both when the cookie changes and each time perses refreshes its requests.

The second commit removed dead code that is no longer used or referenced


web/src/components/dashboards/perses/PersesWrapper.tsx
```diff
@@ -229,52 +226,26 @@ export function PersesPrometheusDatasourceWrapper({
...
-  const datasourceApi = React.useMemo(() => {
-    return new CachedDatasourceAPI(new OcpDatasourceApi(csrfToken, t, PERSES_PROXY_BASE_PATH));
-  }, [csrfToken, t]);
-
-  return (
-    <DatasourceStoreProvider dashboardResource={dashboardResource} datasourceApi={datasourceApi}>
-      <DataQueriesProvider definitions={queries}>{children}</DataQueriesProvider>
-    </DatasourceStoreProvider>
+  const [csrfToken, setCsrfToken] = React.useState(getCSRFToken());
+  const [datasourceApi, setDatasourceApi] = React.useState(
```
jgbernalp, line 232: This probably defeats the purpose of caching the API, in principle the api should not change when the CSRF token changes. Wouldn't it be better to pass the `getCSRFToken` function instead of the actual CSRFToken, this way we can still cache the API but avoid caching the token.

PeterYurkovich:
/unhold
branching is done

PeterYurkovich:
/label ux-approved
/label qe-approved
Bug fix so no UX is needed. QE to be completed post merge

PR #417, opened 2025-05-27
OU-830: fix: add error boundary to variable dropdown

jgbernalp:
This PR:

- adds an error boundary to be able to display variable errors without creating a blank screen
- adds optional chaining to dropdown values as they might be undefined as the type is Partial


jgbernalp:
/label ux-approved
No UX changes
/label qe-approved
QE check will be performed post merge

PR #431, opened 2025-06-13
OU-814: Fix perses dashboard selection

jgbernalp:
fixes [OU-814](https://issues.redhat.com/browse/OU-814)

This PR:

- fixes a typo in `combinedInitialLoad`
- removes `withFallback` HOC to allow the console to use its Error boundary and be able to troubleshoot, as the old implementation just showed a blank screen.
- memoizes the `DashboardSkeleton` for better performance
- Sets the dashboard manually into the Perses dashboard store when the selection changes, as the URL changes only worked for new selections, the URL change is kept to load the initial dashboard if its present in the URL.
- fixes the dashboard name using the `displayName` and falling back to `name` to avoid undefined names

https://github.com/user-attachments/assets/4b46d2a4-47fc-4784-b819-e642aeca2262

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ build-mcp-image:
 	DOCKER_FILE_NAME="Dockerfile.mcp" scripts/build-image.sh
 
 .PHONY: build-dev-mcp-image
-build-mcp-image:
+build-dev-mcp-image:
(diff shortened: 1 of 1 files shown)


jgbernalp:
/label ux-approved
No ux changes

PR #434, opened 2025-06-15
OU-788: [Auto][BVT] - merging to main

etmurasaki:


PeterYurkovich (changes requested):
Could you squash your commits down so we don't include the merge commits into the main branch

PR #451, opened 2025-07-01
OU-880: fix: allow correct selection of dashboard with same name in different projects

jgbernalp:
This PR fixes:

- [OU-880](https://issues.redhat.com//browse/OU-880)
- [OU-879](https://issues.redhat.com//browse/OU-879) 

While fixing the dashboard selection I realized the legacy dashboards were broken and displaying a blank page. This was due to reusing the time range dropdowns from Perses without the correct provider.

In order to simplify the code and make it more easy to read, I refactored the legacy dashboards and perses dashboards to reduce the amount of shared components and removed the PersesContext that was driving the selection, the task related to this is https://issues.redhat.com/browse/OU-879


etmurasaki:
/label qe-approved
- for monitoring-plugin I could test in 4.20 injecting it to CMO and now I could see Dashboards (legacy) being loaded 
correctly
- for monitoring-console-plugin I could test in both 4.19 and 4.20 and Perses dashboards were reloaded correctly even with the same dashboard name presented in different namespaces. The layout was different between both and I could see them being rendered correctly

PR #459, opened 2025-07-10
OU-721: reset showgraph state when unmounting button

PeterYurkovich:


PeterYurkovich:
/label ux-approved
Bugfix PR, no UX needed

PR #465, opened 2025-07-16
OU-594: [Incidents] UI is not embedded in the Alerting section

zhuje:
### JIRA 
https://issues.redhat.com/browse/OU-594

### Screen Capture 
https://github.com/user-attachments/assets/4c2153b2-5b95-43ea-b097-1dbc7cff0625


jgbernalp:
@zhuje @etmurasaki we need to test this also when there is no MCO in the cluster.

web/package.json
```diff
@@ -42,7 +42,7 @@
     "@lezer/highlight": "^1.0.0",
     "@lezer/lr": "^1.0.0",
     "@mui/material": "^6.1.10",
-    "@openshift-console/dynamic-plugin-sdk": "^4.19.0-prerelease.1",
+    "@openshift-console/dynamic-plugin-sdk": "^4.19.0-prerelease.2",
```
PeterYurkovich, line 45: The 4.19.0 version has been fully released. If we are already upgrading the package, we should upgrade to the newest one.

web/src/components/alerting/AlertingPage.tsx
```diff
@@ -29,89 +16,34 @@ const AlertRulesPage = React.lazy(
...
-      navigate(paths[tabNumber]);
-      return;
-    }
-  };
+  const pages = [
+    {
+      href: 'alerts',
+      nameKey: 'Alerts',
+      component: AlertsPage,
+      name: 'Alerts',
```
PeterYurkovich, line 25: Similar to how the console codebase does it, we should include a comment above theses with the t function call, so that the translation service can still find them
```js
      // t('Alerts')
      name: 'Alerts',
```

zhuje:
/hold 

Hold until this PR is merged [NO-JIRA: Convert incidents to typescript ](https://github.com/openshift/monitoring-plugin/pull/470) then rebase.

web/src/components/alerting/AlertingPage.tsx
```diff
@@ -29,89 +16,37 @@ const AlertRulesPage = React.lazy(
...
     <>
       <PageSection hasBodyWrapper={false}>
         <Title headingLevel="h1">{t('Alerting')}</Title>
-        <Tabs activeKey={activeTabKey} onSelect={handleTabClick}>
-          {/* Add href to tab so that when hovering you see the link in the bottom left */}
-          <Tab eventKey={0} title={<TabTitleText>{t('Alerts')}</TabTitleText>} />
-          <Tab eventKey={1} title={<TabTitleText>{t('Silences')}</TabTitleText>} />
-          <Tab eventKey={2} title={<TabTitleText>{t('Alerting rules')}</TabTitleText>} />
-        </Tabs>
+        <HorizontalNav contextId="admin-console-observe" pages={pages} />
```
PeterYurkovich, line 48: Other change for doing perspective specific would be adding the perspective in the contectID here using the usePerspective hook

config/incidents.patch.json
```diff
@@ -3,29 +3,14 @@
     "op": "add",
     "path": "/extensions/1",
     "value": {
-      "type": "console.page/route",
+      "type": "console.tab",
```
PeterYurkovich, line 6: While we are waiting for the more comprehensive solution with including the perspective in the console extension point, can we duplicate the `console.tab` extension and add one for the virtualization perspective as well

config/incidents.patch.json
```diff
@@ -3,29 +3,14 @@
...
-    "op": "add",
-    "path": "/extensions/1",
-    "value": {
-      "type": "console.navigation/href",
-      "flags": {
-        "required": ["PROMETHEUS", "MONITORING", "CAN_GET_NS"]
-      },
-      "properties": {
-        "id": "incidents",
+        "contextId": "admin-console-observe",
```
PeterYurkovich, line 8: we should change the name of the contextId to be something a bit more specific than `{perspective}-console-observe` since that could refer to any of our pages. Maybe something to the extent of `{perspective}-alerts-nav`.

config/incidents.patch.json
zhuje: Added another extension point for virtualization perspective

web/src/components/alerting/AlertingPage.tsx
zhuje: Updated with string interpolation for adding the perspective in the contextID
```
 const [perspective] = useActivePerspective();

  // contextId allow console.tab extensions to be injected
  // https://github.com/openshift/console/blob/main/frontend/packages/console-dynamic-plugin-sdk/docs/console-extensions.md#consoletab
  const contextId = `${perspective}-alerts-nav`;
```

config/incidents.patch.json
zhuje: Updated contextId for console.tabs to include {perspective}-alerts-nav (i.e., admin-alerts-nav, virtualization-perspective-alerts-nav)

etmurasaki:
/label qe-approved
tested after Peter's fixes on:
- Virtualization was showing Incidents tab and it shouldn't
- Alerting pages on Virtualization, ACM were not working correctly
It is working fine

PeterYurkovich:
/label ux-approved
This change was driven by UX

PR #470, opened 2025-07-16
NO-JIRA: Convert incidents to typescript

jgbernalp:


PeterYurkovich:
I think we should target this to main and then backport to 0.5

PR #475, opened 2025-07-23
NO-JIRA: refactor incidents to add typescript

jgbernalp:


jgbernalp:
/label ux-approved
no ux changes

etmurasaki:
As I had commented in https://issues.redhat.com/browse/OU-952, with this code I couldn't see incidents in the charts, while removing it from the env, I could see them.

PR #480, opened 2025-07-23
OU-388: add units selector to the metrics page

PeterYurkovich:
This PR looks to add a units selector to the metrics page and allow links to the metrics page to specify what units they would like to have displayed once clicked.

https://github.com/user-attachments/assets/0a17a733-181e-4203-96d3-3708856ad35e


PeterYurkovich:
cc: @fkargbo

PR #481, opened 2025-07-25
OCPBUGS-59772: fix: redirect alertmanager settings to console owned page

jgbernalp:


jgbernalp:
/label ux-approved
no ux changes

jgbernalp:
The changes on the console side are enough as the specific routes will be matched by the console code in: https://github.com/openshift/console/pull/15325

PR #483, opened 2025-07-30
OU-658

PeterYurkovich:


jgbernalp:
I think a `MonitoringProvider` makes total sense for this. To provide prometheus or alert-manger APIs URLs based on the current user permissions rather than being located in a specific perspective, and provide the right links based on the current perspective. We could create hooks like:

- useMetrics()
- useRules()
- useAlerts()

that pull the values based on configuration, user permission, perspective and cluster.

This will probably centralize the switch cases we have in the components that are specific to perspectives.

web/src/components/alerting/AlertingPage.tsx
```diff
@@ -1,23 +1,51 @@
...
+import {
+  HorizontalNav,
+  NamespaceBar,
+  useActivePerspective,
+} from '@openshift-console/dynamic-plugin-sdk';
+import { MonitoringContext, MonitoringProvider } from '../../contexts/MonitoringContext';
 
-const AlertsPage = React.lazy(
-  () => import(/* webpackChunkName: "AlertsPage" */ '../alerting/AlertsPage'),
+const MpCmoAlertsPage = React.lazy(() =>
```
jgbernalp, line 11: we should check if setting the same chunkName does not bloat the chunk including both pages, better to use different chunkNames.

web/src/components/alerting/AlertingPage.tsx
```diff
@@ -1,23 +1,51 @@
...
+  import(/* webpackChunkName: "AlertsPage" */ './AlertsPage').then((module) => ({
+    default: module.MpCmoAlertsPage,
+  })),
 );
-const SilencesPage = React.lazy(
-  () => import(/* webpackChunkName: "SilencesPage" */ '../alerting/SilencesPage'),
+const McpAcmAlertsPage = React.lazy(() =>
```
jgbernalp, line 16: Maybe the naming should be based on the operator deploying it rather than its functionality as this might easily change in the future, for example adding/removing perspectives.

```suggestion
const CooAlertsPage = React.lazy(() =>
```

web/src/components/alerting/SilenceEditPage.tsx
```diff
@@ -31,15 +33,21 @@ const EditInfo = () => {
   );
 };
 
-export const SilenceEditPage = () => {
+const SilenceEditPage = () => {
   const { t } = useTranslation(process.env.I18N_NAMESPACE);
-  const { silencesKey, perspective } = usePerspective();
+  const { plugin } = useContext(MonitoringContext);
```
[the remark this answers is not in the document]
PeterYurkovich: Sounds good, I'll make this change

web/src/components/Incidents/IncidentsPage.tsx
```diff
@@ -92,28 +91,28 @@ const IncidentsPage = () => {
...
-    state.plugins.mcp.getIn(['incidentsData', 'incidentsActiveFilters']),
+    state.plugins.mcp.get('incidentsData')?.get('incidentsActiveFilters')?.toJS(),
   );
   const incidentGroupId = useSelector((state: MonitoringState) =>
-    state.plugins.mcp.getIn(['incidentsData', 'groupId']),
+    state.plugins.mcp.get('incidentsData')?.get('groupId'),
   );
   const alertsData = useSelector((state: MonitoringState) =>
-    state.plugins.mcp.getIn(['incidentsData', 'alertsData']),
+    state.plugins.mcp.get('incidentsData')?.get('alertsData'),
```
jgbernalp, line 108: in some it seems we have toJS which will create a new object, meaning it might generate re renders when the object has the same data.

PeterYurkovich: Yes, this was likely the root cause of an issue I was running into yesterday, which led me to add the `useDeepMemo` hook. The reason I was trying to not use the `getIn` function is that it is unable to type anything with dynamic keys. For example if we have our store of 

```ts
{
  alerting: {
    [namespace: string]: {
      a: string
      b: int
      c: array<{
        d: int
      }>
    }
  }
}
```

And you run `getIn(['alerting', 'openshift-monitoring'])` or `getIn(['alerting', 'openshift-monitoring', 'a'])` the return values of both will be `never` rather than the appropriate typing of `undefined | X`

With all that said, getIn is the correct solution for the reason you pointed out, I'll just look to manually re-add the types after again.

web/src/components/alerting/AlertsPage.tsx
```diff
@@ -66,7 +71,14 @@ const AlertsPage_: FC = () => {
 
     const clusterArray = Array.from(clusterSet);
     return clusterArray.sort();
-  }, [data]);
+  }, [alerts]);
+
+  const namespacedAlerts = useMemo(() => {
+    if (perspective === 'acm' || namespace === ALL_NAMESPACES_KEY) {
```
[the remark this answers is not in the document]
PeterYurkovich: You're correct. I got too wrapped up in the redux stuff that I lost sight of that. I'll look to adjust the useAlerts hook to return the data as well

PR #487, opened 2025-08-04
OCPBUGS-55455: feat: add alerting rules tab to dev console

jgbernalp:
This PR adds the alerting rule tab to the dev console, to allow filtering of "Not Firing" alerts, which in principle are alerting rules.

### DIFF
--- a/web/console-extensions.json
+++ b/web/console-extensions.json
@@ -171,6 +171,17 @@
     }
   },
   {
+    "type": "console.tab",
+    "properties": {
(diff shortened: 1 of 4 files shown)


PeterYurkovich:
/lgtm

Of note this will mostly be overridden by #483 on the 4.20 branch going forward, but I think it makes sense to merge these changes in and then rebase the other PR ontop of it to keep the same git history for the 4.20-4.X branches

PR #490, opened 2025-08-05
OU-960: data-test injection into src code

etmurasaki:


web/src/components/data-test.ts
```diff
@@ -0,0 +1,193 @@
+export const AlertsPage = {
```
jgbernalp, line 1: To avoid aliasing imports like `import { AlertsPage as test} ...` we can use a more appropriate name:

```suggestion
export const AlertsPageTestIds = {
```

web/src/components/data-test.ts
```diff
@@ -0,0 +1,193 @@
...
+  ExpireXSilencesButton: 'expire-x-silences-button',
+  SilencesTable: {
+    SilencesTable: 'silences-table',
+    SelectAllCheckbox: 'select-all-checkbox',
+    SelectCheckbox: 'select-checkbox',
+    SilenceResourceIcon: 'silence-resource-icon',
+    SilenceResourceLink: 'silence-resource-link',
+    SilenceMatchersList: 'silence-matchers-list',
+    SilenceCount: 'silence-count',
+    SilenceState: 'silence-state',
```
jgbernalp, line 85: This id is duplicated in other page, I think the purpose of this file is to reuse these ids. We could use a flat structure and create a single object that holds the unique ids. Otherwise it will become difficult to know from which id is this `silence-state`

PR #494, opened 2025-08-07
OCPBUGS-60198: fix legend height to avoid scroll overlap

jgbernalp:


jgbernalp:
/label ux-approved
no ux changes

PR #504, opened 2025-08-14
NO-JIRA: update to expect NodeJS v22+

zhuje:


zhuje:
/label ux-approved 
No ux required; only README changes

/label qe-approved
No qe required; only README changes

PR #507, opened 2025-08-19
NO-JIRA: README.md update

zhuje:


zhuje:
/label ux-approved 
No ux changes

/label qe-approved 
No necessary, only README updates

PR #509, opened 2025-08-21
OBSINTA-776: Incident detection tests

DavidRajnoha:
**Draft PR: Incident Detection E2E Tests Refactor and Expansion**

This Pull Request introduces initial UI tests for the incident detection feature. Its primary purpose is to serve as a proof of concept, a basic smoke test, and to demonstrate the capabilities for incident page interaction within the testing framework.

### Key Changes:

* **Initial Incident Detection E2E Tests**

    * `views/incidents-page.ts` includes **Cypher selectors**, which facilitate operations on the incident page elements.

    * `e2e/incidents/01.incidents.cy.ts` incorporates tests designed to verify the setup of the incidents plugin and to interact with the various elements present on the page. This test is dependent on existing alerts to be present in the system at the moment.

    * `e2e/incidents/00.incidents.cy.ts` is truly e2e test withou external dependencies, it deploys a pod configured to fail upon startup, thereby generating a `KubePodCrashLoop` alert. The test subsequently verifies that this alert appears within the system.

    * **Error handling has been implemented in `index.ts`** to catch the 'ResizeObserver loop limit' exception, which is otherwise triggered by the incident page.

    * `support/commands.ts`  includes a command for **deploying failing pods**, along with the necessary label additions to ensure proper incident detection functionality during setup.

    * `fixtures/incidents/` contains the **YAML files required for the failing pod deployment**, including a custom alert rule.

This commit lays a foundational groundwork for further automated testing of the incident detection system. It is marked as a draft, I would like to hear feedback on it to know that this is the right direction to take.

### DIFF
--- a/web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
+++ b/web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
@@ -1,0 +1,59 @@
+/*
+The test verifies the whole lifecycle of the Incident feature, without any external dependencies.
+*/
+import { commonPages } from '../../views/common';
+import { incidentsPage } from '../../views/incidents-page';
+
+// Set constants for the operators that need to be installed for tests.
+const MCP = {
+  namespace: 'openshift-cluster-observability-operator',
+  packageName: 'cluster-observability-operator',
+  operatorName: 'Cluster Observability Operator',
+  config: {
+    kind: 'UIPlugin',
+    name: 'monitoring',
+  },
+};
+
+const MP = {
+  namespace: 'openshift-monitoring',
+  operatorName: 'Cluster Monitoring Operator',
+};
+
+describe('Incidents', () => {
+  before(() => {
+    cy.afterBlockCOO(MCP, MP); // Following cypher best practices, the cleanup is done before the test block
+    cy.beforeBlockCOO(MCP, MP);
+    cy.createKubePodCrashLoopingAlert();
+  });
+
+  after(() => {
+    cy.afterBlockCOO(MCP, MP); // For compatibility with other tests
+  });
+
+  it('Admin Perspective - Incidents tab renders and responds to interactions', () => {
+    cy.log('1.1 Navigate to Alerting → Incidents');
+    incidentsPage.goTo();
+    commonPages.titleShouldHaveText('Incidents');
+  });
+
+  it('Incident with KubePodCrashLooping alert is present in the alerts table', () => {
+    incidentsPage.goTo();
+    commonPages.titleShouldHaveText('Incidents');
+    incidentsPage.clearAllFilters();
+    
+    const intervalMs = 60_000;
+    const maxMinutes = 20; 
+
+    cy.waitUntil(() => incidentsPage.findIncidentWithAlert('KubePodCrashLooping'), { 
+      interval: intervalMs, 
+      timeout: maxMinutes * intervalMs 
+    });
+
+    incidentsPage
+      .elements
+      .alertsTable()
+      .contains('KubePodCrashLooping')
+      .should('exist');
+  });
+});
--- a/web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
+++ b/web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
@@ -1,5 +1,6 @@
 /*
 The test verifies the whole lifecycle of the Incident feature, without any external dependencies.
+The run time can be 15 - 20 minutes. (Waiting untill the incident detection captures the new alert)
 */
 import { commonPages } from '../../views/common';
 import { incidentsPage } from '../../views/incidents-page';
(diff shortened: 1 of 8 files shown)


web/cypress/views/incidents-page.ts
```diff
@@ -0,0 +1,130 @@
...
+      });
+  },
+
+  // Centralized element selectors - all selectors defined in one place
+  elements: {
+    // Page structure
+    toolbar: () => cy.get('#toolbar-with-filter'),
+
+    // Controls and filters
+    daysSelect: () => cy.get('button[data-ouia-component-id="OUIA-Generated-MenuToggle-4"]'),
```
etmurasaki, line 112: just to mention it seems a bit fragile, but if you tested and it always ends -4, go ahead. other way would be something like this: button[data-ouia-component-id^="OUIA-Generated-MenuToggle".

etmurasaki (commented):
/lgtm

web/cypress/views/incidents-page.ts
sthundat: @DavidRajnoha you could ask dev team to create a unique  id for this element like `[data-testid="days-select"]`

web/cypress/views/incidents-page.ts
```diff
@@ -0,0 +1,130 @@
...
+    filtersSelect: () => cy.get('button[data-ouia-component-id="OUIA-Generated-MenuToggle-5"]'),
```
sthundat, line 113: use more stable selector

web/cypress/e2e/incidents/01.incidents.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+  beforeEach(() => {
+
+    cy.log('Navigate to Observe → Incidents');
+    incidentsPage.goTo();
+    commonPages.titleShouldHaveText('Incidents');
+  });
+
+  it('renders toolbar and toggle charts button', () => {
+    incidentsPage.elements.toolbar().should('be.visible');
+    incidentsPage.elements.toggleChartsButton().should('be.visible');
```
sthundat, line 54: Instead of checking the visibility you can click the button and assert the page content .Modify the test name appropriately

web/cypress/e2e/incidents/01.incidents.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+  it('toggles Critical filter and updates the URL', () => {
+    incidentsPage.clearAllFilters();
+    incidentsPage.toggleFilter('Critical');
+    cy.url().should('match', /incidentFilters=.*Critical/);
+  });
+
+  it('shows charts and alerts empty state initially', () => {
+    incidentsPage.elements.incidentsChartTitle().should('be.visible');
+    incidentsPage.elements.alertsChartTitle().should('be.visible');
+    incidentsPage.elements.alertsChartEmptyState().should('exist');
```
sthundat, line 72: here you can check the 'Hide graph|Show graph' empty state behavior as well

web/cypress/views/incidents-page.ts
```diff
@@ -0,0 +1,130 @@
...
+    clearAllFiltersButton: () => cy.contains('button', 'Clear all filters'),
+    toggleChartsButton: () => cy.contains('button', /Hide graph|Show graph/),
+
+    // Charts and visualizations  
+    incidentsChartTitle: () => cy.contains('Incidents Timeline'),
+    incidentsChartCard: () => cy.contains('Incidents Timeline').closest('.pf-v6-c-card'),
```
sthundat, line 119: Could we add a data-testid  like `data-testid="incidents-chart-card"`

web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
```diff
@@ -0,0 +1,59 @@
...
+    name: 'monitoring',
+  },
+};
+
+const MP = {
+  namespace: 'openshift-monitoring',
+  operatorName: 'Cluster Monitoring Operator',
+};
+
+describe('Incidents', () => {
```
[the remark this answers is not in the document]
falox: IMO it's ok, the name is the same of the folder and it's following the pattern used by other e2e tests https://github.com/openshift/monitoring-plugin/blob/main/web/cypress/e2e/monitoring/01.bvt_monitoring.cy.ts#L28

Maybe the `BVT:` prefix if missing, but please double check with the other authors.

web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
```diff
@@ -0,0 +1,59 @@
...
+    cy.afterBlockCOO(MCP, MP); // Following cypher best practices, the cleanup is done before the test block
+    cy.beforeBlockCOO(MCP, MP);
+    cy.createKubePodCrashLoopingAlert();
+  });
+
+  after(() => {
+    cy.afterBlockCOO(MCP, MP); // For compatibility with other tests
+  });
+
+  it('Admin Perspective - Incidents tab renders and responds to interactions', () => {
```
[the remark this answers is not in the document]
falox: I would do the opposite: keep this, that clearly tests a specific aspect and remove the line 42 in the second test, that is effectively redundant.

web/cypress/e2e/incidents/01.incidents.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+  namespace: 'openshift-monitoring',
+  operatorName: 'Cluster Monitoring Operator',
+};
+
+const ALERTNAME = 'Watchdog';
+const NAMESPACE = 'openshift-monitoring';
+const SEVERITY = 'Critical';
+const ALERT_DESC = 'This is an alert meant to ensure that the entire alerting pipeline is functional. This alert is always firing, therefore it should always be firing in Alertmanager and always fire against a receiver. There are integrations with various notification mechanisms that send a notification when this alert is not firing. For example the "DeadMansSnitch" integration in PagerDuty.'
+const ALERT_SUMMARY = 'An alert that should always be firing to certify that Alertmanager is working properly.'
+describe('Incidents', () => {
```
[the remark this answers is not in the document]
falox: As before, I think the feature name "Incidents" is fine. The detailed descriptions are in the tests themselves. Double check the pattern to use

web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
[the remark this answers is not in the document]
sthundat: @etmurasaki Could you please confirm if we have any naming conventions in place for test files and their corresponding describe blocks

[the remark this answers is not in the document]
DavidRajnoha: I will update the descriptions, I missed that they are the same. The difference between the tests is the setup. This test deploys its own crashing pod and checks for the alert. It truly tests the e2e behavior of the incidents. The second is currently relying on some alert existing in the system (so the bars would be there) and in the future I would like to set up mocking / intercepting requests so the alerts are there. Its aim is therefore solely the verification of the incident UI component.

web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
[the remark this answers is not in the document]
DavidRajnoha: Based on the [cypress best practices](https://docs.cypress.io/app/core-concepts/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion ), I would go the Savitha's route. On the other hand, the first test is helping me now when developing as a feedback that the basic setup works and the testing is waiting for the long running test which waits on the alert creation.

web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
```diff
@@ -0,0 +1,59 @@
...
+    cy.log('1.1 Navigate to Alerting → Incidents');
+    incidentsPage.goTo();
+    commonPages.titleShouldHaveText('Incidents');
+  });
+
+  it('Incident with KubePodCrashLooping alert is present in the alerts table', () => {
```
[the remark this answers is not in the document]
DavidRajnoha: The name of the alert is the key element that is checked here, to verify that the particular alert is displayed in the incidents. I am thinking of taking it further and customizing the name of the alert and checking that this particular name is present in the alerts.

Doing the alert based on filtering is for sure possible, but I would prefer to add it in the next PR to keep the work more incremental. I am also thinking whether it makes sense to add them here or whether they would be a better fit in the other test which is working more on the UI display, as this one is focused on the backend / UI connection.

web/cypress/views/incidents-page.ts
DavidRajnoha: I will focus on after this gets merged and make a PR that adds the selectors to the UI and uses them in the tests.

web/cypress/views/incidents-page.ts
DavidRajnoha: I will focus on after this gets merged and make a PR that adds the selectors to the UI and uses them in the tests.

web/cypress/views/incidents-page.ts
DavidRajnoha: I will focus on after this gets merged and make a PR that adds the selectors to the UI and uses them in the tests.

DavidRajnoha (commented):
Addressed some comments, the not-addressed now are taken in account and will be addressed in the changes in the code.

web/cypress/e2e/incidents/00.coo_incidents_e2e.cy.ts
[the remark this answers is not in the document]
etmurasaki: Hi there, I had BVT aka Build Verification Test just to check the environment/cluster/our component is not broken, testable with the very basic flow. REG as regression, it is exercising more actions, to be run in other job and it can take longer. It made sense for me to break bc I have many pages to test.
For Incidents, since it is only one page, it is fine to maintain in only one file, but it is totally up to you! We can adjust accordingly later.

PR #510, opened 2025-08-21
 NO-JIRA: Test config interactive setup

DavidRajnoha:
The script will interactively ask you for CYPRESS_ env variables.
When run using `$ source ./configure-env.sh`, it automatically exports them (Bash only),
`./configure-env.sh` creates file you can source yourself by `source export-env.sh`

It searches for all potential kubeconfig file in $HOME/Downloads, so you do not have to look for the path.
It has been mostly generated with Claude and Cursor.


DavidRajnoha:
> @DavidRajnoha if you could also take advantage of this PR to document on https://github.com/openshift/monitoring-plugin/blob/main/web/cypress/README.md file, would be great!

Added the documentation in Readme

etmurasaki (approved):

PR #511, opened 2025-08-21
OU-781: fix links in incidents table

Fewwy:
I moved the logic for adding rules data to the data we used for mapping the table to the IncidentsPage component
That way, we don't have a race condition and don't need to wait for the rules data to resolve before we map over the table data and populate the table.

The table now has only links to the alerting rules, not alerting details.
```
How I tested this
1. crc config set enable-cluster-monitoring true
2. crc start
3. upload mock incidents using https://github.com/openshift/cluster-health-analyzer
In the root of cluster-healt-analyzer run:
A) oc apply -f manifests/backend -f manifests/frontend
B) go run ./main.go simulate
C) promtool tsdb create-blocks-from openmetrics cluster-health-analyzer-openmetrics.txt

D) for d in data/*; do
 echo $d
 kubectl cp $d openshift-monitoring/prometheus-k8s-0:/prometheus -c prometheus
done

The "prometheus-k8s-0" in the last command could have a different name. Open your local dev environment to check the name

I run a monitoring plugin using the instructions in the README that enable features:
# In a separate terminal
$ make start-feature-console

# In a separate terminal
$ make start-feature-backend
```

### DIFF
--- a/web/src/components/Incidents/IncidentsDetailsRowTable.tsx
+++ b/web/src/components/Incidents/IncidentsDetailsRowTable.tsx
@@ -8,7 +8,7 @@ import {
 import { BellIcon } from '@patternfly/react-icons';
 import { Bullseye, Spinner } from '@patternfly/react-core';
 import { Link } from 'react-router-dom';
-import { AlertResource } from '../utils';
+import { RuleResource } from '../utils';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


falox:
/retitle OU-781: fix links in incidents table

Fewwy:
@jgbernalp It was approved, can we merge this?

PR #513, opened 2025-08-26
OU-769: implement a custom tooltip for incidents

jgbernalp:
This PR adds a custom tooltip so there are no issues in safari and the style matches better other tooltips in monitoring.

Safari:
Screenshot 2025-08-26 at 11 29 55

Chrome:
Screenshot 2025-08-26 at 11 30 04


jgbernalp:
/label ux-approved
no UX changes
/label qe-approved
QE can check post merge

falox:
@jgbernalp That's great! Now the tooltip looks much better. Could you apply the same changes to AlertsChart.tsx? It's still using the old tooltip

PR #520, opened 2025-09-03
RHINENG-20424: Incident ID filter for the Incidents Page

Fewwy:
Mocks:
https://issues.redhat.com/browse/RHINENG-20424
Incident ID is a filter for the group_id parameter of the Incidents

1) Added groupId to the incidentActiveFilters in state management, now it works the same way as state/severity filters
2) Added a filter for incident ID, which is populated with Incident IDs
3) Updated the IncidentsChart component to make it work with the new incidentsActiveFilters.groupId, removed state management parts for the old groupId
4) I removed the logic for the opacity change in the Incidents chart; now, we hide other bars when we filter by Incident.
5) Added a function onFilterToggle for toggling the expansion of filters 
6) Added necessary logic for updating the URL params, clearing the filters by single deletion or group deletion
7) Clear all filters now, do not clear the "days" selection, previously it was always refreshing it back to 7 days.

Small UX improvements:
1) Removed "Time range" above the days selector because it is pushing the entire container down and making it look bad. This change was agreed upon with Alberto
2) Removed "gap" for PageSection component so it wouldn't create an empty space between filter chips and "hide/show" chart button
3) Remove the filterIcon component for the IncidentFilterToolbarItem
4) Fixed the "gap" between filter type filter and the filters by using "is hidden" css

### DIFF
--- a/web/src/actions/observe.ts
+++ b/web/src/actions/observe.ts
@@ -32,7 +32,6 @@ export enum ActionType {
   ShowGraphs = 'showGraphs',
   SetIncidents = 'setIncidents',
   SetIncidentsActiveFilters = 'setIncidentsActiveFilters',
-  SetChooseIncident = 'setChooseIncident',
   SetAlertsData = 'setAlertsData',
   SetAlertsTableData = 'setAlertsTableData',
   SetAlertsAreLoading = 'setAlertsAreLoading',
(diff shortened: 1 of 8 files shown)


falox:
Thanks @Fewwy, here is a first round of things to fix:

Bugs:
- Alerts and component details are not expanded on incident selection
- The incident selection is not preserved if you change the time range

UI bugs:
- The space between filter category and filter values is dynamically changing on category change. It should be always the same. If possible the width of the category dropdown should be always the same (i.e. the max width, the one of incident id item)

UX inconsistencies:
- The incident list as checkbox list is wrong, since it's a single selection. You can hide the checkbox or transform to radio button list

PR #521, opened 2025-09-04
OBSINTA-798: Cypress Incident Tests – Query Mocking

DavidRajnoha:
The PR Implements mocking of requests from incident page to prometheuse, in particular the requests for cluster:health:components:map and for ALERTS. It enables running targeted UI tests independent of the live alerts in the cluster, which are often impossible to set up in a way to replicate the desired scenarios.

The mocks are defined using YAML-based fixture scenarios. This commit includes 5 scenarios that set up different UI states.

Custom Cursor command is added, which can be used to generate those YAML scenarios from text or screenshot. Reliability is supported by including a schema verification script that the AI uses to checks its created artifact.

web/cypress/e2e/incidents/02.incidents-mocking-example.cy.ts showcases the mocking states. The test does not contain any assertions, in the future PRs, I want to add cypress tags to the testsuite that will disable running this test by default.

The mocking is integrated into the web/cypress/e2e/incidents/01.incidents.cy.ts to be independent of the cluster state.

New env variable CYPRESS_TIMEZONE is added, it is needed for the mocked alert timelines to align with the time in your cluster. It is included in the configure-env.sh script.


DavidRajnoha:
The mocking works by intercepting the request containing cluster:health:components:map prometheus query and returning response with the incidents in the same format:

[mock_response_cluster:health:components:map.txt](https://github.com/user-attachments/files/22168291/mock_response_cluster.health.components.map.txt)

After an incident bar is clicked, frontend makes two other types of requests, one for the particular incident and the second for corresponding alerts. Those are also mocked and data just for the filtered data are returned:
[mock_response_cluster:health:components:map_filtered.txt](https://github.com/user-attachments/files/22168294/mock_response_cluster.health.components.map_filtered.txt)
[mock_response_alerts_query.txt](https://github.com/user-attachments/files/22168293/mock_response_alerts_query.txt)

sthundat:
@DavidRajnoha Can you attach the test results also here?

web/cypress/support/incidents_prometheus_query_mocks/prometheus-mocks.ts
```diff
@@ -0,0 +1,81 @@
...
+  cy.intercept('GET', '/api/prometheus/api/v1/query_range*', (req) => {
+    const url = new URL(req.url, window.location.origin);
+    const query = url.searchParams.get('query') || '';
+
+    console.log(`INTERCEPTED: ${req.method} ${req.url}`);
+    console.log(`Query: ${query}`);
+
+    let results: any[]; 
+
+    if (query.includes('cluster:health:components:map')) {
```
sthundat, line 29: you can improve Readability with Early Return ,It's  clear what happens when the query isn't handled, and then the valid cases are processed afterward. Example
`if (!query.includes('cluster:health:components:map') && !query.includes('ALERTS{')) {
  console.log(`Passing through non-mocked query`);
  req.continue();
  return;
}

const results = query.includes('cluster:health:components:map')
  ? createIncidentMock(incidents, query)
  : createAlertDetailsMock(incidents, query);`

web/cypress/fixtures/incident-scenarios/0-healthy-cluster.yaml
sthundat, line 1: What is the purpose of the watchdog alert on the Incident page?
If I remember correctly, it doesn't appear in the incident timeline.
If you want to simulate an informational alert, it would be better to add a simple alert with info severity instead.

DavidRajnoha:
@etmurasaki @IshwarKanse @sthundat I have addressed the points that you have raised. Is there anything preventing this PR to get merged?

IshwarKanse (approved):

PR #535, opened 2025-09-17
OU-863: fix step calculation to fix non existent data gaps

jgbernalp:


etmurasaki:
post-merge testing performed and comments in jira.

PR #540, opened 2025-09-18
OU-1005: upgrade perses plugins

jgbernalp:
This PR:

- Upgrades Perses dependencies to latest v0.52.0 
- Include all the perses code plugins and update them to their latest versions

This is an intermediate step to update the plugins. It will be replaced when we can load the installed plugins dynamically: https://issues.redhat.com/browse/OU-885 
cc @zhuje


zhuje:
/label ux-approved 

No UX changes

PR #542, opened 2025-09-19
NO-JIRA: fix: immer dependency anr patternfly icons import

jgbernalp:
This PR:

- downgrades the immer dependency to use v9.0.21, which is the same used for perses. This removes the warning when loading the app
- Fixes other warnings related to the patternfly icons imports


jgbernalp:
/label ux-approved
no ux changes

etmurasaki:
/label qe-approved
regression testing performed locally, successfully.

PR #545, opened 2025-09-19
OBSINTA-795: Data Test Selectors for Incidents Page

DavidRajnoha:
This PR Adds data-test ids for the incident page, that are used for better stability of the tests and for robustness against future changes. 

A test is added into `01.incidents.cy.ts` that uses the same alert traversal functionality as `00.coo_incidents_e2e.cy.ts`, thus decouples the identification of potential UI issues from e2e issues.


Fewwy (approved):
Changes for the Incidents page look fine. LGTM

PR #549, opened 2025-09-22
OBSINTA-789: Fix alerts chart empty state

Fewwy:
1)Memoized clickHandler function to avoid recreation of the function.
2) Moved clickHandler onClick function to the parent component and used combination of useCallback and React.memo to avoid additional rerenders.
3)Added conditions to the Table and Alerts chart that ensures that empty state will be shown when there no incidents selected. That fixes the bug/regression.

### DIFF
--- a/web/src/components/Incidents/AlertsChart/AlertsChart.tsx
+++ b/web/src/components/Incidents/AlertsChart/AlertsChart.tsx
@@ -30,6 +30,7 @@ import { useTranslation } from 'react-i18next';
 import { AlertsChartBar } from '../model';
 import { setAlertsAreLoading } from '../../../store/actions';
 import { MonitoringState } from '../../../store/store';
+import { isEmpty } from 'lodash-es';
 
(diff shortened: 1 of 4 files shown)


web/src/components/Incidents/AlertsChart/AlertsChart.tsx
```diff
@@ -30,6 +30,7 @@ import { useTranslation } from 'react-i18next';
 import { AlertsChartBar } from '../model';
 import { setAlertsAreLoading } from '../../../store/actions';
 import { MonitoringState } from '../../../store/store';
+import * as _ from 'lodash-es';
```
jgbernalp, line 33: we should avoid importing the whole library and use only the functions needed. Even if this already present on the code we are moving away from it

Fewwy: Replaced it with only isEmpty, force pushed the update

PR #560, opened 2025-09-28
OU-1003: automation monitoring stable

etmurasaki:


etmurasaki:
closing since it will be incorporated into OU-995.

PR #569, opened 2025-10-02
OBSINTA-789: Fix alert loading and processing

falox:
This PR rewrites the alert loading logic and fixes the following bugs: OU-1014, OU-1020, OU-1021. 
Details in the individual commit messages

Assisted-by: Claude


DavidRajnoha:
The PR does not fully fix OU-1014 – the case with just single data point is still not rendered correctly. The fix does not have to be included here, but please do not close the JIRA if this gets merged.

PR #570, opened 2025-10-02
OU-1024: Add translations for all strings in Incidents page

DavidRajnoha:
Translations were only partially applied for the incidents page. This PR wrap all strings in the code with the translation function. For the translations to be fully working, the files in `/web/locales/` still need to be updated.

Translation helpers are added for the PatternFly chip deletion callbacks in `ToolbarItemFilter.tsx`. The callbacks give us the displayed translated value, so the  reverseTranslateLabel is used to translate it back so the correct value can be passed to the onDeleteIncidentFilterChip call.

Assisted by Cursor with claude-4.5-sonnet model.


DavidRajnoha:
@tremes @falox @rioloc

PR #573, opened 2025-10-06
OU-1016: avoid displaying stale alerts data when selecting another incident detail

tremes:


web/src/components/Incidents/AlertsChart/AlertsChart.tsx
```diff
@@ -50,19 +50,38 @@ const AlertsChart = ({ theme }: { theme: 'light' | 'dark' }) => {
...
+      return [];
+    }
+
+    return alertsData;
+  }, [alertsData, alertsAreLoading, selectedGroupId]);
+
   // Use dynamic date range based on actual alerts data instead of fixed chartDays
   const dateValues = useMemo(() => {
-    if (!Array.isArray(alertsData) || alertsData.length === 0) {
+    if (!Array.isArray(displayAlertsData) || displayAlertsData.length === 0) {
```
jgbernalp, line 74: it seems displayAlertsData is always an array, would be better to use the correct type when defining the const

Fewwy:
Don't merge this, there is a bug. If you open filters dropdown and then deselect selected incident it will refresh the page and the filter will be still open 
image

DavidRajnoha:
@Fewwy I saw this too, but it should be fixed on the latest changes when @tremes addressed this in particular

PR #580, opened 2025-10-07
OCPBUGS-62276: allow additional TLS config

jgbernalp:
This PR adds additional configuration options to be set from environment or flags:

- tls-min-version
- tls-max-version
- tls-cipher-suites
- tls-curves


cmd/plugin-backend.go
```diff
@@ -95,3 +123,107 @@ func mergeEnvValueInt(key string, arg int, defaultValue int) int {
...
+	log.Warnf("Invalid TLS version %q, using default VersionTLS12", version)
+	return tls.VersionTLS12
+}
+
+func parseCipherSuites(ciphers string) []uint16 {
+	if ciphers == "" {
+		return nil
+	}
+
+	cipherMap := map[string]uint16{
```
PeterYurkovich, line 146: Is there a standardized way that Red Hat is performing these parse checks? My concern is that the issue calls out "Hardcoding TLS configuration [...] does not adhere to our evolving, centrally-managed security policy[...]" and that by hardcoding the possible values when a change (like a new addition) is made to that centrally managed security policy we will then need to update this to account for it

jgbernalp: It seems the ciphers can be listed dynamically but not the other values. For now I extracted them into functions so they are easy to maintain

simonpasquier: IIUC the second leg of the change will in CMO which will enforce the TLS settings defined at the cluster level (it does it already for all other components that it manages).

PR #603, opened 2025-10-15
OU-991: monitoring testings refactoring for reusability for admin and kubevirt

etmurasaki:
In order to have same Monitoring testing running against different perspectives, in this case, Virtualization, existing Monitoring scenarios under e2e folder were moved to support folder and exported in a way that in e2e testing files we can call them out just changing the perspective in beforeEach() part.


web/cypress/README.md
```diff
@@ -1,6 +1,9 @@
 # Openshift Monitoring Plugin and Monitoring Console Plugin UI Tests
 These console tests are related to Monitoring Plugin deployed by Cluster Monitoring Operator (CMO) as part of OCP - Observe menu with Alerting, Metrics, Dashboards pages and other related Alerting links.
-Besides, Monitoring Console Plugin deployed by Cluster Observability Operator through Monitoring UIPlugin installation. 
+Besides, Monitoring Console Plugin deployed by Cluster Observability Operator through Monitoring UIPlugin installation.
+
+## Test Documentation
+For a comprehensive overview of all E2E test scenarios, including COO (Cluster Observability Operator), Monitoring, and Incidents tests, see [E2E_TEST_SCENARIOS.md](./E2E_TEST_SCENARIOS.md).
```
DavidRajnoha: Can we discuss this on the next QE meeting? We have been also thinking about the inclusion of the test scenarios, but we have in mind (and locally implemented) a bit different approach (gitlab, due to the possibility of containing customer info in the bug description that might be there). It would be worth considering whether we would want to have this approach unified or not. For sure, I would like to prevent keeping information about incident scenarios at two separate places.

PR #633, opened 2025-11-05
OBSINTA-858: configurable Cypress COO namespace and idempotent cleanup 

DavidRajnoha:
Contains fixes that were necessary to make the testsuite integrated in the release pipeline work.

Add CYPRESS_COO_NAMESPACE environment variable to enable flexible namespace configuration for Cluster Observability Operator installations. This allows testing with different namespace configurations (e.g., 'coo' instead of the default 'openshift-cluster-observability-operator').

Additionally, it makes the cleanup delete commands not fail when the deleted resources is not present, which is expected as they are run before the tests to ensure clean environment.

Changes:
- Add CYPRESS_COO_NAMESPACE env var with default value 'openshift-cluster-observability-operator'
- Update cypress.config.ts to read the new environment variable
- Update configure-env.sh to prompt for and export namespace configuration
- Update coo_stage.sh installation script to use configurable namespace
- Update all test files to use Cypress.env('COO_NAMESPACE') instead of hardcoded values
- Update mock generators to use the env var for namespace
- Add --ignore-not-found flag to all oc delete commands in cleanup functions
- Update dashboard management to use sed for on-the-fly namespace substitution
- Add documentation in README.md for the new configuration option
Added:
- Fix oc label commands to use singular 'namespace' instead of 'namespaces'
- Remove --ignore-not-found flag from cluster-admin role removal
- Add test-cypress-incidents-regression npm script

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "monitoring-plugin",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {}
+}
--- a/web/cypress/support/commands/operator-commands.ts
+++ b/web/cypress/support/commands/operator-commands.ts
@@ -192,7 +192,7 @@ const operatorUtils = {
         'Operator installed successfully',
       );
       cy.exec(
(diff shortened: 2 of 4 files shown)


DavidRajnoha:
There are few fixes that I realized I have not introduced and I would like to merge together.

PR #634, opened 2025-11-06
COO-1311: Fix incorrect tooltip positioning for silenced alerts

falox:
The redundant "Silenced: true" line was causing incorrect tooltip positioning


DavidRajnoha:
label /qe-approved

PR #637, opened 2025-11-06
[release-4.17] OU-1058: Don't show TP button in non-administrator

PeterYurkovich:
This PR looks to add a perspective key to non-admin perspective alert charts so that the TP injects only into the admin perspective.


PeterYurkovich:
/label ux-approved
No UX changes

PR #642, opened 2025-11-12
NO-JIRA: add devspace setup

PeterYurkovich:


PeterYurkovich:
/label ux-approved
/label qe-approved
Development only changes so no QE or UX review needed

PR #643, opened 2025-11-12
OU-877: perses e2e testing

etmurasaki:
E2E testing for Perses in a way to be reusable in different perspectives.
- e2e folder contains a new folder: perses and a cypress file calling out testing scenarios that are exported and located under support folder, following recommendations from claude code;
- this way the same testing scenarios can be called out just changing the before, beforeEach changing the perspective;
- this approach will be applied to Monitoring testing in other PR.
- how to run will be written as part of Testing for monitoring-plugin repo as a whole, but this cypress testing also takes into consideration exported CYPRESS variables to install or skip installation of COO, to be decoupled of any other COO test.
- uninstallation is not part of this test due to CI/CD approach to reuse sessions and skip installation, to speed up execution time. A separated uninstallation testing will be developed.

image

### DIFF
--- a/web/cypress/support/commands/utility-commands.ts
+++ b/web/cypress/support/commands/utility-commands.ts
@@ -1,3 +1,4 @@
+import { Classes, DataTestIDs, LegacyTestIDs } from "../../../src/components/data-test";
 export {};
 
 declare global {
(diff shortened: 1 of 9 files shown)


web/cypress/support/commands/utility-commands.ts
```diff
@@ -57,20 +57,38 @@ Cypress.Commands.add('waitUntilWithCustomTimeout', (
...
-      if ($element.attr('data-checked-state') !== 'true') {
-        cy.byTestID('showSystemSwitch').siblings('span').eq(0).should('be.visible').click();
+    cy.wait(2000);
+    cy.get('body').then(($body) => {
+      const hasNamespaceBarDropdown = $body.find('[data-test-id="namespace-bar-dropdown"]').length > 0;
+      if (hasNamespaceBarDropdown) {
+        cy.byLegacyTestID('namespace-bar-dropdown').find('button').scrollIntoView().should('be.visible');
+        cy.byLegacyTestID('namespace-bar-dropdown').find('button').scrollIntoView().should('be.visible').click({force: true});
+      } else {
+        cy.byClass('pf-v6-c-menu-toggle co-namespace-dropdown__menu-toggle').scrollIntoView().should('be.visible');
```
jgbernalp: maybe we can remove the pf specific class matcher or add a new test id so the tests work when backported.

PR #657, opened 2025-11-25
Add base alert management API

machadovilaca:


simonpasquier (commented):
Can we have a separate PR on main for the golangci-lint addition? It's very much appreciated but it'd be easier to review separately. Can we also have a minimal golangci-lint configuration? Once this is done we should also consider adding a job to openshift/release to enforce the verification.
Regarding goimports, I'd recommend using the `gci` linter which does the same IIUC.

.github/workflows/unit-tests.yaml
simonpasquier: you can't have GitHub actions on github.com/openshift

PR #661, opened 2025-12-01
NO-JIRA: use tanstack-query for dashboard config fetching

PeterYurkovich:
An example usage of the useQuery for fetching data in the monitoring-plugin. I'd like to take this PR as a place to discuss if this is something we want to move forward with, and if so how we should structure our codebase differently. In the perses codebase they have a set of "client" files which contain the fetch and hook functions, with the data returned being used in their components. Since our team doesn't have control over the backend, we can't structure the API endpoints to return the data as we need it, so we may need a third functional layer there which performs the data formatting, because (as seen in this PR) we may want to adjust or filter the data without needing to refetch. Let me know what you think

### DIFF
--- a/web/src/components/dashboards/legacy/legacy-dashboard-page.tsx
+++ b/web/src/components/dashboards/legacy/legacy-dashboard-page.tsx
@@ -19,9 +19,8 @@ type LegacyDashboardsPageProps = {
 const LegacyDashboardsPage_: FC<LegacyDashboardsPageProps> = ({ urlBoard }) => {
   const { project, setProject } = useOpenshiftProject();
   const {
-    legacyDashboardsError,
+    legacyDashboards,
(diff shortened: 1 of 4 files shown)


jgbernalp:
I think this is valuable as is moving all the useEffects and useState handling to tanstack. For other complex cases, like alerting, I think is valuable to refactor our logic to abstract the data fetching logic including filtering and formatting into hooks, as is it is demonstrated in this PR with legacy dashboards. This will allow us to use directly fetch, tanstack or any other without the components actually caring about it.

Regarding the file structure, it makes sense to put the hook near the components that are using it. This will also apply better when other teams are contributing and are responsible for a specific folder.

Some topics to have in mind:

- The new bundle size and initial load impact
- Stale results / cache invalidation
- Request cancellation

PR #673, opened 2025-12-09
OU-1082: claude slash commands cypress-setup and cypress-run

etmurasaki:
Creation of 2 claude slash commands:
`/cypress-setup` and `/cypress-run`

When running `/cypress-setup`, it validates the pre-reqs and opens a new terminal to prepare it with all cypress variables needed, such as base URL, credentials, kubeconfig etc.
Then `/cypress-run` will print to you all available commands for you to choose it and it will run in the terminal previously opened. 

This way, will make claude available to perform other things instead of running and controlling the testing execution.

### DIFF
--- a/.claude/commands/cypress/cypress-setup.md
+++ b/.claude/commands/cypress/cypress-setup.md
@@ -1,0 +1,75 @@
+---
+name: cypress-setup
+description: Automated Cypress environment setup with interactive configuration
+parameters: []
+---
+
+# Cypress Environment Setup
+
+This command sets up the Cypress testing environment by checking prerequisites, installing dependencies, and interactively configuring environment variables.
+
+## Instructions for Claude
+
+Follow these steps in order, always:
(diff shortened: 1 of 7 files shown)


.claude/commands/cypress/cypress-setup.md
```diff
@@ -0,0 +1,383 @@
...
+parameters: []
+---
+
+# Cypress Setup (Automated)
+
+This command validates prerequisites, detects workspace, creates setup script, and opens terminal automatically - all with minimal approvals.
+
+---
+
+## Execution Flow
```
jgbernalp: For my this failed as it was trying to execute the command in the claude code terminal, instead of opening a new one. Maybe it will be better to be more explicit that a new terminal should be opened and the script should be executed there. Also maybe we don't need to embeed the whole script in the command but in a different folder and point the command to it.

PR #674, opened 2025-12-10
OU-632: Split ALERTS query_range into several requests for Incidents

rioloc:
This PR introduces the following:

- Simplify Prometheus query logic removing not needed `group_left` 
- Split single _ALERTS_ query into several chunk of fixed sizes to avoid url length errors

Assisted-By: Claude Code


falox:
/retitle [OU-632](https://issues.redhat.com//browse/OU-632): Split ALERTS query_range into several requests for Incidents

PR #677, opened 2025-12-10
OU-1113: handle all ns selected for tenancy user

PeterYurkovich:
There are infrequent times (such as when a user first logs in) when a dev user can access the all namespaces selection in the namespaces dropdown. In those cases we remove our special handling of the namespace parameter


etmurasaki:
/override pull-ci-openshift-monitoring-plugin-main-e2e-monitoring

PR #713, opened 2026-01-02
NO-JIRA: feat: use esbuild-loader, add push manifest command to makefile

jgbernalp:
This PR:

- Uses esbuild-loader to speed up prod builds and reduce memory consumption
- Adds a command to publish the manifest created when building cross platform
- Increases default nodejs memory for building


etmurasaki:
/override ci/prow/e2e-monitoring

flaky validation will be removed to not conflict in claimed cluster from pool and cluster provisioned from scratch, impacting in when inline appears.

PR #724, opened 2026-01-08
OU-1159: [Edit Dashboards] - Edit List Variable - Not able to erase Custom All Value

zhuje:
[OU-1159](https://issues.redhat.com/browse/OU-1159) [Edit Dashboards] - Edit List Variable - Not able to erase Custom All Value

Upgrading to  "@perses-dev/plugin-system": "0.53.0-rc.1" because this version already contains the fix for this issue https://github.com/perses/shared/blob/main/plugin-system/src/components/Variables/VariableEditorForm/VariableEditorForm.tsx#L332


zhuje:
Closing because after QE testing, the bug still exists and needs to be fixed upstream.

PR #729, opened 2026-01-12
NO-JIRA: [release-4.21] fix: remove unnecessary package.lock

jgbernalp:


PeterYurkovich:
I just realized that we found the same dangling file at the same time and both went to delete it, if you want to just use the cherry pick from the other PR thats fine too

PR #731, opened 2026-01-14
NO-JIRA: monitoring testing - metrics - ungraphable for periodics

etmurasaki:
image

### DIFF
--- a/web/cypress/support/monitoring/02.reg_metrics_2.cy.ts
+++ b/web/cypress/support/monitoring/02.reg_metrics_2.cy.ts
@@ -265,23 +265,8 @@ export function testMetricsRegression2(perspective: PerspectiveConfig) {
 
   });
 
-  it(`${perspective.name} perspective - Metrics > Ungraphable results`, () => {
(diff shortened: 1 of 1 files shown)


jgbernalp:
@etmurasaki can you elaborate why clicking on more elements on the predefined queries will fix the flakiness? and why are we testing this "ungraphable" scenario?

etmurasaki:
@jgbernalp this ungraphable scenario is part of alternative paths such as invalid query to expose different states: No datapoints, No datapoints with alert danger icon. 

This scenario is not passing in periodic jobs when a cluster is provisioned from scratch, while in presubmit it is reusing / claiming a cluster from the pool, always with data. 

Per your recommendation and I agree, we should avoid relying on data we do not control, but it was a tentative to add more predefined queries to reach this ungraphable state in periodic - that is still rendering the graph probably due to the low amount of data, and in presubmit first 6 predefined ones already reach it, so adding more won't hurt for presubmit, but probably work in periodic. 

Let me know if you are ok to delete this scenario from the test suite for metrics. Thank you!

PR #736, opened 2026-01-20
COO-1515: add the cluster-health-analyzer feature

tremes:
This is adding a new `cluster-health-analyzer` feature so that the incident page is still displayed when this feature is enabled. It's related to the https://github.com/rhobs/observability-operator/pull/976


jgbernalp:
This might create the problem that 2 incidents tabs are added if incidents and health analyzer are both active. I think a better approach would be to keep the `config/cluster-health-analyzer.patch.json` remove the old incidents patch, having a single patch. And add logic to the go server to apply the patch in the case incidents or health analyzer features are enabled. Something similar to what we did in the logging plugin: https://github.com/openshift/logging-view-plugin/blob/main/pkg/server/manifest_handler.go#L44

jgbernalp:
/label qe-approved
tested the extension injection for both features

PR #749, opened 2026-01-30
OU-1040: feat/absolute start dates

rioloc:
## Problem
`Start` dates in tooltips are relatives to the numer of selected days as time span.

For example, if today is 30 Jan 2026 and an incident started on 22 Jan 2026, the `Start` date will be displayed as:

- 29 Jan 2026 if "Last 1 Day" is selected
- 27 Jan 2026 if "Last 3 Days" is selected
- 23 Jan 2026 if "Last 7 Days" is selected
- 22 Jan 2026 if "Last 15 Days" is selected. (The only correct one)

## Fix
The absolute start date of an incident/alert is always displayed, and it is not related to the number of selected days.

## Solution
Absolute timestamps for `cluster_health_components_map{}`, for incidents, and `ALERTS{}` for alerts are retrieved by performing an instant query call to Prometheus in order to get the `min_over_time(timestamp(cluster_health_components_map{}))`, which return the timestamp of the first datapoint for that metric. (Same for ALERTS).
The result is saved into redux store and then used to match related incident/alert in order to  update the Start date displayed in the tooltip.

## Before
[main.webm](https://github.com/user-attachments/assets/61097ffb-44fd-4f55-8b55-36ec6d90df77)

## After
[feat⁄absolute-start-dates.webm](https://github.com/user-attachments/assets/1919b0c2-6ec7-4e5e-a716-c7792092e5f1)

## Summary by CodeRabbit

* **New Features**
  * Alerts and incidents are grouped so identical alerts share rows and severity segments are split for clearer status tracking.
  * Charts and tables now use a 15-day window for richer historical context.

* **Improvements**
  * More accurate start/end timestamps and 5-minute-aligned rounding for charts and tables.
  * Trailing padding removed from intermediate severity segments to prevent visual overlap.

### DIFF
--- a/web/src/components/Incidents/IncidentsDetailsRowTable.tsx
+++ b/web/src/components/Incidents/IncidentsDetailsRowTable.tsx
@@ -25,10 +25,8 @@ const IncidentsDetailsRowTable = ({ alerts }: IncidentsDetailsRowTableProps) => 
     if (alerts && alerts.length > 0) {
       return [...alerts]
         .sort((a: IncidentsDetailsAlert, b: IncidentsDetailsAlert) => {
-          const aFirstTimestamp = a.firstTimestamps[0][1];
-          const bFirstTimestamp = b.firstTimestamps[0][1];
-          const aStart = aFirstTimestamp > 0 ? aFirstTimestamp : a.alertsStartFiring;
-          const bStart = bFirstTimestamp > 0 ? bFirstTimestamp : b.alertsStartFiring;
+          const aStart = a.firstTimestamp > 0 ? a.firstTimestamp : a.alertsStartFiring;
+          const bStart = b.firstTimestamp > 0 ? b.firstTimestamp : b.alertsStartFiring;
           return aStart - bStart;
         })
         .map((alertDetails: IncidentsDetailsAlert, rowIndex) => {
@@ -50,8 +48,8 @@ const IncidentsDetailsRowTable = ({ alerts }: IncidentsDetailsRowTableProps) => 
               <Td dataLabel="expanded-details-firingstart">
                 <Timestamp
                   timestamp={
-                    (alertDetails.firstTimestamps[0][1] > 0
-                      ? alertDetails.firstTimestamps[0][1]
+                    (alertDetails.firstTimestamp > 0
+                      ? alertDetails.firstTimestamp
                       : alertDetails.alertsStartFiring) * 1000
                   }
                 />
--- a/web/src/components/Incidents/IncidentsPage.tsx
+++ b/web/src/components/Incidents/IncidentsPage.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable react-hooks/exhaustive-deps */
 import { useMemo, useState, useEffect, useCallback } from 'react';
 import { useSafeFetch } from '../console/utils/safe-fetch-hook';
-import { createAlertsQuery, fetchDataForIncidentsAndAlerts, fetchInstantData } from './api';
+import { createAlertsQuery, fetchDataForIncidentsAndAlerts } from './api';
 import { useTranslation } from 'react-i18next';
 import {
   Bullseye,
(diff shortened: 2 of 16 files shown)


web/src/components/Incidents/IncidentsDetailsRowTable.tsx
```diff
@@ -45,13 +47,27 @@ const IncidentsDetailsRowTable = ({ alerts }: IncidentsDetailsRowTableProps) =>
...
+                    ) * 1000
+                  }
+                />
               </Td>
               <Td dataLabel="expanded-details-firingend">
                 {!alertDetails.resolved ? (
                   '---'
                 ) : (
-                  <Timestamp timestamp={alertDetails.alertsEndFiring * 1000} />
+                  <Timestamp
```
PeterYurkovich: Why are we rounding the start time but not the end time?

web/src/components/Incidents/IncidentsPage.tsx
```diff
@@ -229,49 +235,74 @@ const IncidentsPage = () => {
...
-          );
+    // Guard: don't process if no incidents selected or timeRanges not ready
+    if (incidentForAlertProcessing.length === 0 || timeRanges.length === 0) {
+      return;
+    }
+
+    const currentTime = incidentsLastRefreshTime;
+
+    // Fetch timestamps and alerts in parallel, but wait for both before processing
+    const timestampsPromise = Promise.all(
```
PeterYurkovich: The timestamps are being fetched a forced 15 days while the the incidents and alerts are still fetched based on the timeRanges (duration) selected. Is there a reason for this disconnect of the time ranges being searched?

web/src/components/Incidents/IncidentsPage.tsx
```diff
@@ -294,14 +325,34 @@ const IncidentsPage = () => {
       ? `cluster_health_components_map{group_id='${selectedGroupId}'}`
       : 'cluster_health_components_map';
 
-    Promise.all(
+    // Fetch timestamps and incidents in parallel, but wait for both before processing
+    const timestampsPromise = Promise.all(
+      ['min_over_time(timestamp(cluster_health_components_map)[15d:5m])'].map(async (query) => {
```
PeterYurkovich: same here

web/src/components/Incidents/IncidentsTable.tsx
```diff
@@ -180,7 +181,9 @@ export const IncidentsTable = () => {
                       )}
                     </Td>
                     <Td dataLabel={columnNames.startDate}>
-                      <Timestamp timestamp={getMinStartDate(alert) * 1000} />
+                      <Timestamp
+                        timestamp={roundTimestampToFiveMinutes(getMinStartDate(alert)) * 1000}
```
PeterYurkovich: I see nearly every location that the start date is displayed that it is rounded down at that location to be a 5 minute timestamp. Could we just round the timestamp down before storing it in the store and then remove all of theses `roundTimestampToFiveMinutes` calls for just the one?

web/src/components/Incidents/IncidentsDetailsRowTable.tsx
rioloc: The new start time is taken from the `timestamp` metrics we are retrieving with the new calls to Prometheus. This is taken as an absolute value and then rounded (this is done only to give an user experience coherent with the end time).
The end time is still displayed with the previous way which already relied on some rounding logic.

web/src/components/Incidents/IncidentsPage.tsx
```diff
@@ -229,49 +235,74 @@ const IncidentsPage = () => {
...
+        }
+        if (!isEmpty(filteredData)) {
+          dispatch(setAlertsAreLoading({ alertsAreLoading: false }));
+        } else {
+          dispatch(setAlertsAreLoading({ alertsAreLoading: true }));
+        }
+      })
+      .catch((err) => {
+        // eslint-disable-next-line no-console
+        console.log(err);
```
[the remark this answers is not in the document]
rioloc: Fixed in https://github.com/openshift/monitoring-plugin/pull/749/changes/0c541366e2a5f6289a802b4a9278c1ff7450dc99#diff-bcb1f04d349a97a7b87b8855f505f2d5f207843cb1a2420893060e9526c47361L303

web/src/components/Incidents/IncidentsTable.tsx
rioloc: Good point. Fixed in https://github.com/openshift/monitoring-plugin/pull/749/commits/0c541366e2a5f6289a802b4a9278c1ff7450dc99

PR #754, opened 2026-02-05
OU-1062: feat: mark alert menu active for incidents tab

jgbernalp:
Screenshot 2026-02-05 at 13 24 35

### DIFF
--- a/web/console-extensions.json
+++ b/web/console-extensions.json
@@ -7,7 +7,11 @@
       "href": "/monitoring/alerts",
       "perspective": "admin",
       "section": "observe",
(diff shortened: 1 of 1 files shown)


jgbernalp:
@DavidRajnoha I see the error is related to a not found application launcher:
```

  1) BVT: Incidents - e2e
       "before all" hook for "1. Admin perspective - Incidents page - Incident with custom alert lifecycle":
     AssertionError: Timed out retrying after 80000ms: Expected to find element: `[data-test-id="application-launcher"]`, but never found it.
     ```
Is this a version specific case?

PR #755, opened 2026-02-06
refactor: Move Go code to dedicated backend/ directory

machadovilaca:


DavidRajnoha:
The proposed structure is aligned with what I described on the call.

There seems to be some issue with the PR tests now, I will retrigger them later if needed, their pass is neccessary to ensure that the change is reflected in our automation.

PR #813, opened 2026-03-04
OU-1260: [release-4.16] feat: use npm instead of yarn

jgbernalp:


jgbernalp:
needs: https://github.com/openshift/release/pull/77710

PR #841, opened 2026-03-12
CNV-80608: add CRD support and create alert rule API

sradco:
Add AlertingRule, AlertRelabelConfig, and
RelabeledRules CRD interfaces with the
management client, router, server wiring,
and POST /api/v1/alerting/rules endpoint.


pkg/management/management.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/openshift/monitoring-plugin/pkg/k8s"
+)
+
+type client struct {
+	k8sClient k8s.Client
+}
+
+func (c *client) IsPlatformAlertRule(prId types.NamespacedName) bool {
+	return c.k8sClient.Namespace().IsClusterMonitoringNamespace(prId.Namespace)
```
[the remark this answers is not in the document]
sradco: CMO's platform Prometheus evaluates every PrometheusRule in a `openshift.io/cluster-monitoring=true` namespace regardless of who created it. The namespace IS the routing boundary.

go.mod
```diff
@@ -1,67 +1,79 @@
 module github.com/openshift/monitoring-plugin
 
-go 1.24.0
+go 1.25.0
```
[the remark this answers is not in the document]
sradco: Fixed

pkg/k8s/relabeled_rules.go
```diff
@@ -0,0 +1,447 @@
...
+
+	if !cache.WaitForNamedCacheSync("RelabeledRulesConfig informer", ctx.Done(),
+		rrm.prometheusRulesInformer.HasSynced,
+		rrm.secretInformer.HasSynced,
+	) {
+		return nil, fmt.Errorf("failed to sync RelabeledRulesConfig informer")
+	}
+
+	go rrm.worker(ctx)
+	rrm.queue.Add("initial-sync")
```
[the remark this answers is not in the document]
sradco: fixed

pkg/k8s/relabeled_rules.go
```diff
@@ -0,0 +1,447 @@
...
+				if arName := alertingRuleOwner(promRule); arName != "" {
+					rule.Labels[managementlabels.AlertingRuleLabelName] = arName
+				}
+
+				ruleManagedBy, relabelConfigManagedBy := rrm.determineManagedBy(ctx, promRule, alertRuleId)
+				if ruleManagedBy != "" {
+					rule.Labels[managementlabels.RuleManagedByLabel] = ruleManagedBy
+				}
+				if relabelConfigManagedBy != "" {
+					rule.Labels[managementlabels.RelabelConfigManagedByLabel] = relabelConfigManagedBy
```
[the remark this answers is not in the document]
sradco: Fixed

pkg/k8s/relabeled_rules.go
```diff
@@ -0,0 +1,447 @@
...
+func (rrm *relabeledRulesManager) List(ctx context.Context) []monitoringv1.Rule {
+	rrm.mu.RLock()
+	defer rrm.mu.RUnlock()
+
+	var result []monitoringv1.Rule
+	for _, rule := range rrm.relabeledRules {
+		result = append(result, rule)
+	}
+
+	return result
```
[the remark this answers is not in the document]
sradco: Fixed

pkg/management/alert_rule_preconditions.go
```diff
@@ -0,0 +1,98 @@
...
+		if gitOpsManaged, _ := k8s.IsExternallyManagedObject(pr); gitOpsManaged {
+			return notAllowedGitOpsEdit()
+		}
+	}
+	// ARC metadata GitOps block
+	if arc != nil && k8s.IsManagedByGitOps(arc.Annotations, arc.Labels) {
+		return notAllowedGitOpsEdit()
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
sradco: Fixed

pkg/management/testutils/k8s_client_mock.go
```diff
@@ -0,0 +1,419 @@
...
+
+// MockClient is a mock implementation of k8s.Client interface
+type MockClient struct {
+	TestConnectionFunc      func(ctx context.Context) error
+	PrometheusRulesFunc     func() k8s.PrometheusRuleInterface
+	AlertRelabelConfigsFunc func() k8s.AlertRelabelConfigInterface
+	AlertingRulesFunc       func() k8s.AlertingRuleInterface
+	RelabeledRulesFunc      func() k8s.RelabeledRulesInterface
+	NamespaceFunc           func() k8s.NamespaceInterface
+}
```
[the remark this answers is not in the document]
sradco: Fixed

pkg/server.go
```diff
@@ -146,7 +150,23 @@ func createHTTPServer(ctx context.Context, cfg *Config) (*http.Server, error) {
...
+			return nil, fmt.Errorf("failed to create k8s client for alert management API: %w", err)
+		}
+
+		if err := k8sClient.TestConnection(ctx); err != nil {
+			return nil, fmt.Errorf("failed to connect to kubernetes cluster for alert management API: %w", err)
+		}
+
+		managementClient = management.New(ctx, k8sClient)
+		log.Info("alert management API enabled")
+	}
```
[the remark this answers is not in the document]
sradco: Fixed

sradco:
Hi @jgbernalp , @jan--f , @simonpasquier, all issues are fixed. Please review.

internal/managementrouter/create_alert_rule.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	if err != nil {
+		handleError(w, err)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusCreated)
+	_ = json.NewEncoder(w).Encode(CreateAlertRuleResponse{Id: id})
```
jgbernalp: Can this fail?, if so should we return an error or at least log it?

internal/managementrouter/router.go
```diff
@@ -0,0 +1,61 @@
...
+	r.HandleFunc("/api/v1/alerting/rules", httpRouter.CreateAlertRule).Methods(http.MethodPost)
+
+	return r
+}
+
+func writeError(w http.ResponseWriter, statusCode int, message string) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(statusCode)
+	resp, _ := json.Marshal(map[string]string{"error": message})
+	_, _ = w.Write(resp)
```
jgbernalp: same here it seems errors are being ignored

pkg/k8s/relabeled_rules.go
```diff
@@ -0,0 +1,468 @@
...
+// DetermineManagedBy determines the managed-by labels for a PrometheusRule alert rule.
+func DetermineManagedBy(ctx context.Context, alertRelabelConfigs AlertRelabelConfigInterface, namespaceManager NamespaceInterface, promRule *monitoringv1.PrometheusRule, alertRuleId string) (string, string) {
+	rrm := &relabeledRulesManager{
+		alertRelabelConfigs: alertRelabelConfigs,
+		namespaceManager:    namespaceManager,
+	}
+	return rrm.determineManagedBy(ctx, promRule, alertRuleId)
+}
+
+func (rrm *relabeledRulesManager) List(ctx context.Context) []monitoringv1.Rule {
```
jgbernalp, line 466: Do we need a way to paginate this list, could this be an issue with clusters with many relabeledRules?

internal/managementrouter/create_alert_rule.go
sradco: fixed

internal/managementrouter/router.go
sradco: fixed

pkg/k8s/relabeled_rules.go
sradco: The cache stores one small struct per alerting rule. 
Even 1000 rules is only a few MB. 
Changes are batched into a single rebuild. 
In later PRs, HTTP responses are filtered server-side (namespace, name, source, labels) before returning, so pagination isn't needed. Responses are already bounded subsets of the cache.

jgbernalp: But the filters would be optional selected from the frontend right? I'm fine with this changes, but we need to consider that the current rules support `groupNextToken` to paginate using a cursor approach. Maybe it can be added in a follow up PR

sradco: +1. I will add this in follow up PR.

pkg/management/create_user_defined_alert_rule.go
```diff
@@ -0,0 +1,134 @@
...
+	alertrule "github.com/openshift/monitoring-plugin/pkg/alert_rule"
+	"github.com/openshift/monitoring-plugin/pkg/k8s"
+	"github.com/openshift/monitoring-plugin/pkg/managementlabels"
+)
+
+const (
+	DefaultGroupName = "user-defined-rules"
+)
+
+func (c *client) CreateUserDefinedAlertRule(ctx context.Context, alertRule monitoringv1.Rule, prOptions PrometheusRuleOptions) (string, error) {
```
jan--f, line 19: Any reason we don't validate the alert here like in `CreatePlatformAlertRule`? https://github.com/openshift/monitoring-plugin/pull/841/changes#diff-40b75231a364caa6afd3cc4babac7fc0cbde2de0aa5d0dbbf7d0007bd511b21fR22?

sradco: You are right. I now extracted a shared `validateAlertRuleInputs` helper that both `CreatePlatformAlertRule` and `CreateUserDefinedAlertRule` calls, since the core checks (alert name, expr, severity) apply to both paths.

sradco:
@jan--f @jgbernalp, fixed the issues you raised. Thank you for the review.

pkg/server.go
```diff
@@ -146,7 +149,29 @@ func createHTTPServer(ctx context.Context, cfg *Config) (*http.Server, error) {
...
+		k8sClient, err := k8s.NewClient(initCtx, k8sconfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create k8s client for alert management API: %w", err)
+		}
+
+		if err := k8sClient.TestConnection(initCtx); err != nil {
+			return nil, fmt.Errorf("failed to connect to kubernetes cluster for alert management API: %w", err)
+		}
+
+		managementClient = management.New(ctx, k8sClient)
```
jgbernalp, line 170: Nothing that needs to be fixed i this PR, but have we consider how this might operate in HA? e.g. having the plugin in a deployment with 2 or more replicas

pkg/managementlabels/management_labels.go
```diff
@@ -0,0 +1,20 @@
+package managementlabels
+
+const (
```
simonpasquier, line 3: (nit) I'd appreciate having a clear explanation for each const value.

internal/managementrouter/router.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return r
+}
+
+func writeError(w http.ResponseWriter, statusCode int, message string) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(statusCode)
+	resp, err := json.Marshal(map[string]string{"error": message})
+	if err != nil {
+		log.Printf("failed to marshal error response: %v", err)
```
simonpasquier: is this even possible? shouldn't we panic() instead?

internal/managementrouter/create_alert_rule.go
```diff
@@ -0,0 +1,61 @@
...
+type CreateAlertRuleRequest struct {
+	AlertingRule   *monitoringv1.Rule                `json:"alertingRule,omitempty"`
+	PrometheusRule *management.PrometheusRuleOptions `json:"prometheusRule,omitempty"`
+}
+
+type CreateAlertRuleResponse struct {
+	Id string `json:"id"`
+}
+
+const maxRequestBodyBytes = 1 << 20 // 1 MB
```
simonpasquier: (nit) it should probably be defined in the server directly if we really care.

internal/managementrouter/create_alert_rule.go
```diff
@@ -0,0 +1,61 @@
...
+	"log"
+	"net/http"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+
+	"github.com/openshift/monitoring-plugin/pkg/management"
+)
+
+type CreateAlertRuleRequest struct {
+	AlertingRule   *monitoringv1.Rule                `json:"alertingRule,omitempty"`
```
simonpasquier: hmm why use the definition from the Prometheus operator here?

pkg/alert_rule/alert_rule.go
```diff
@@ -0,0 +1,86 @@
...
+	"sort"
+	"strings"
+	"unicode/utf8"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+
+	"github.com/openshift/monitoring-plugin/pkg/managementlabels"
+)
+
+var promLabelNameRegexp = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
```
simonpasquier, line 18: Prometheus for the latest OCP (4.22) release should support any UTF-8 label names.

pkg/alert_rule/alert_rule.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Generate SHA256 hash
+	hash := sha256.Sum256([]byte(canonicalPayload))
+
+	return "rid_" + base64.RawURLEncoding.EncodeToString(hash[:])
+}
+
+func NormalizeExpr(expr string) string {
+	// Collapse consecutive whitespace so cosmetic formatting changes do not churn ids.
+	return strings.Join(strings.Fields(strings.TrimSpace(expr)), " ")
```
simonpasquier: Does it mean that `up{job="job 1"}` and `up{job="job       1"}` would be considered the same?

pkg/k8s/alert_relabel_config.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return arcm, nil
+}
+
+func alertRelabelConfigListWatchForAllNamespaces(clientset *osmv1client.Clientset) *cache.ListWatch {
+	return cache.NewListWatchFromClient(clientset.MonitoringV1().RESTClient(), "alertrelabelconfigs", "", fields.Everything())
+}
+
+func (arcm *alertRelabelConfigManager) List(ctx context.Context, namespace string) ([]osmv1.AlertRelabelConfig, error) {
+	arcs := arcm.arcInformer.GetStore().List()
```
simonpasquier, line 50: I'm puzzled here: if we pull the resources from the informer, it means that we can't ensure RBAC? E.g. a resource which the user doesn't have access to might leak to them.

simonpasquier (commented):
## OpenAPI or not

Instead of crafting the API definitions and validations by hand, wouldn't it make sense to use an OpenAPI framework?

## Logging

I see that we use `github.com/sirupsen/logrus` and `log`. We need to pick one library (one important criteria is that it plays nice with k8s client-go) and settle with it. 

## RBAC

I left a comment on the use of informers. I'm still concerned that we might leak information which shouldn't be visible to end-users.

## Testing

I'm not a huge fan of ginkgo tests for unit tests (the Go standard testing framework is good enough IMHO). But in any case, the current test coverage is pretty thin.

Irrespective of this remark what's our testing strategy for this project? I would consider that end-to-end testing is the most beneficial approach: mocking the different interfaces in unit tests is going to be brittle.

pkg/alert_rule/alert_rule.go
sradco: @simonpasquier  Can COO support older versions of OCP? Should we add this protection for them?

pkg/k8s/alert_relabel_config.go
sradco: You are correct. I will fix this to use impersonation.

pkg/alert_rule/alert_rule.go
sradco: +1. I will fix this

sradco:
> ## OpenAPI or not
> Instead of crafting the API definitions and validations by hand, wouldn't it make sense to use an OpenAPI framework?

Will work to address this.
> 
> ## Logging
> I see that we use `github.com/sirupsen/logrus` and `log`. We need to pick one library (one important criteria is that it plays nice with k8s client-go) and settle with it.

True. I will update to use logrus.
> 
> ## RBAC
> I left a comment on the use of informers. I'm still concerned that we might leak information which shouldn't be visible to end-users.

I agree, I will update the code to use impersonation.
> 
> ## Testing
> I'm not a huge fan of ginkgo tests for unit tests (the Go standard testing framework is good enough IMHO). But in any case, the current test coverage is pretty thin.

+1. Will move to stdlib and will add coverage.
> 
> Irrespective of this remark what's our testing strategy for this project? I would consider that end-to-end testing is the most beneficial approach: mocking the different interfaces in unit tests is going to be brittle.

internal/managementrouter/create_alert_rule.go
sradco: +1. Once Ill move the code to the  OpenAPI spec-first approach this will also be fixed.

internal/managementrouter/router.go
sradco: +1 . I will  removing the error check since the input is always safe.

sradco:
@simonpasquier @jgbernalp  thank you for reviewing the PR. 
I updated it based on your comments.
I still have an open question about https://github.com/openshift/monitoring-plugin/pull/841#discussion_r3137742954.
Other than that I believe I addressed all the review comments.

pkg/alert_rule/alert_rule.go
simonpasquier: Yes COO can support multiple minor versions of OpenShift. My assumption is that OCP >= 4.22 versions should support UTF-8 label names.

pkg/k8s/relabeled_rules.go
```diff
@@ -0,0 +1,509 @@
...
+	out := make([]rune, 0, len(s))
+	for _, r := range s {
+		if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') || r == '-' {
+			out = append(out, r)
+		} else {
+			out = append(out, '-')
+		}
+	}
+	// collapse multiple '-' and trim
+	res := strings.Trim(strings.ReplaceAll(string(out), "--", "-"), "-")
```
jgbernalp: This only collapses pairs: `a---b` -> `a--b`, we can use a regex instead

pkg/k8s/prometheus_rule.go
jgbernalp: Should this use also the `clientSetForCtx`?

pkg/management/label_utils.go
```diff
@@ -0,0 +1,12 @@
...
+
+var validSeverities = map[string]bool{
+	"critical": true,
+	"warning":  true,
+	"info":     true,
+	"none":     true,
+}
+
+func isValidSeverity(s string) bool {
+	return validSeverities[s]
```
jgbernalp: should we convert this to lowercase before searching? `Critical` `WARNING` will fail, do we want them to fail?

sradco: Fixed

pkg/k8s/prometheus_rule.go
sradco: Fixed

pkg/k8s/relabeled_rules.go
sradco: Fixed

.github/workflows/unit-tests.yaml
jgbernalp, line 1: This file should be removed as we the OpenShift org does not run Github actions, all is run in prow and go tests are already there. We can remove it in a follow up PR

PR #863, opened 2026-03-27
OU-1175: OCP 4.22 Breaking Changes

PeterYurkovich:
This PR looks to update the monitoring-plugin to handle the 4.22 breaking changes in the console, comprising of shared package updates. The following package updates are the primary focus of this PR:
- @openshift-console/dynamic-plugin-sdk (and the others)
- react
- react-dom
- react-redux
- react-router
- eslint
- typescript
- webpack

Due to an issue with `@openshift-console/dynamic-plugin-sdk@4.22-prerelease.2` the `patch-package` module has been added with a single patch file to adjust one of the shared modules to enable building. This package and the patch file should be removed when the offending function has been fixed in a future `@openshift-console/dynamic-plugin-sdk` release.

There should be no visible changes from this update


web/src/components/dashboards/legacy/useOpenshiftProject.ts
```diff
@@ -0,0 +1,71 @@
...
+import { useMonitoring } from '../../../hooks/useMonitoring';
+import { useParams } from 'react-router';
+import { dashboardsPatchVariable } from '../../../store/actions';
+
+export const useOpenshiftProject = () => {
+  const { perspective } = usePerspective();
+  const [activeNamespace, setActiveNamespace] = useActiveNamespace();
+  const { ns: routeNamespace } = useParams<{ ns?: string }>();
+  const [openshiftProject, setOpenshiftProject] = useQueryParam(
+    QueryParams.OpenshiftProject,
```
jgbernalp, line 17: Are there links defined with this parameter? It seems we are re introducing yet another hook for getting the active namespace/project. We probably should adjust the existing one `useMonitoringNamespace`, and if necessary configure it via props. Otherwise becomes confusing to reason when to use which. WDYT?

web/src/components/dashboards/perses/hooks/useDashboardsData.ts
```diff
@@ -39,8 +40,8 @@ export const useDashboardsData = () => {
     return true;
   }, [persesProjectsLoading, persesDashboardsLoading, initialPageLoad, setInitialPageLoadFalse]);
 
-  const prevDashboardsRef = useRef<DashboardResource[]>([]);
-  const prevMetadataRef = useRef<CombinedDashboardMetadata[]>([]);
+  const [prevDashboards, setPreviousDashboards] = useState<DashboardResource[]>([]);
```
jgbernalp: This was made on purpose to avoid unnecessary re renders. Is there any reason we are re adding this?

web/src/components/dashboards/perses/project/ProjectBar.tsx
```diff
@@ -1,30 +1,24 @@
...
         <ProjectDropdown
           onSelect={(event, newProject) => {
             const params = new URLSearchParams();
-            if (newProject === '') {
-              setActiveProject(null);
-            } else {
-              params.set('project', newProject);
-              setActiveProject(newProject);
-            }
+            params.set('project', newProject);
```
jgbernalp: We should probably use they parameter key from an enum.

web/src/components/MetricsPage.tsx
```diff
@@ -1134,13 +1133,16 @@ const QueryBrowserWrapper: FC<{
...
-    const newParams: Record<string, string> = {};
-    _.each(queryStrings, (q, i) => (newParams[`query${i}`] = q || ''));
+    if (isFirstRender) {
+      return;
+    }
+    const newParams = new URLSearchParams(queryParams);
+    queryStrings.forEach((query, i) => newParams.set(`query${i}`, query || ''));
     if (customDataSourceName) {
-      newParams.datasource = customDataSourceName;
+      newParams.set('datasource', customDataSourceName);
```
jgbernalp: they key here should also come from an enum to avoid desync.

web/src/components/query-browser.tsx
```diff
@@ -640,7 +641,7 @@ const QueryBrowser_: FC<QueryBrowserProps> = ({
   const [samples, setSamples] = useState(maxSamplesForSpan);
   const [updating, setUpdating] = useState(true);
   // Track if we ever received valid data to prevent flickering "No datapoints" during refresh
-  const hasReceivedData = useRef(false);
+  const [hasReceivedData, , setReceivedData] = useBoolean(false);
```
jgbernalp: Using the `useBooelan` creates re renders if the value changes, if a re render is not intended we should keep the ref.

web/package.json
```diff
@@ -44,7 +44,8 @@
     "test-cypress-fast": "node --max-old-space-size=4096 ./node_modules/.bin/cypress run --browser chrome --headless --env grepTags='@smoke --@slow --@demo --@flaky --@xfail'",
     "test-cypress-perses-dev": "node --max-old-space-size=4096 ./node_modules/.bin/cypress run --browser chrome --headless --env grepTags='@perses-dev --@demo --@xfail'",
     "test-cypress-perses": "node --max-old-space-size=4096 ./node_modules/.bin/cypress run --browser chrome --headless --env grepTags='@perses --@smoke --@flaky --@demo --@xfail'",
-    "ts-node": "ts-node -O '{\"module\":\"commonjs\"}'"
+    "ts-node": "ts-node -O '{\"module\":\"commonjs\"}'",
+    "postinstall": "node ./node_modules/.bin/patch-package"
```
jgbernalp: This won't be executed when productizing. Should we patch the SDK or ask the console team to prioritize and wait for the SDK release?

web/src/components/dashboards/legacy/useOpenshiftProject.ts
PeterYurkovich: The `useMonitoringNamespace` is very simple compared to the logic in the logic in the `useOpenshiftProject`, and adding configuration will cause the `useMonitoringNamespace` to become not easily understood for the simple usecase. I could change this to be named something similar to `useDashboardsNamespace` so that it is clear that it's use is only for the dashboard pages. 

As far as if there are links with that parameter, the answer generally is no (unless someone bookmarks it), as the change to have the admin perspective hold the namespaced dashboards only happened recently. However the namespace and the dashboard project MUST be separate query parameters. I pinged you in a thread on slack with a number of different images explaining why it has to be the case. If you have any questions lmk or I can re-link the thread as I'm sure you have a lot of notifications currently

web/package.json
PeterYurkovich: The `Dockerfile.art` (and all other Dockerfiles in our repo) have been updated to include performing the patch, however you are correct that COO won't perform the patch unless we make the change in that Dockerfile as well. Hopefully the change from the SDK team will be completed by then, however it is not the end of the world if it isn't. Since the bug is in the SDK, it only causes an issue during build time, as we use the shared SDK module while running in the console, so there is no code breakage after it is built. There is an open bug ticket for them to address it, and once they do we will revert this part. They have estimated the next SDK build to be around branching time

https://redhat.atlassian.net/browse/OCPBUGS-81319

web/src/components/dashboards/perses/hooks/useDashboardsData.ts
PeterYurkovich: I will look to address the remainder in my next commit, which may take some time due to the complexity of the state interactions that I am working through

PeterYurkovich: Looking back on this, this change was due to the upgrade in eslint config. The two usages of ref's that you point out are fine for usage, so what I'll do is add a file level eslint override for them and revert to using refs. 

In the future as we tighten our lint rules we can determine what the best way to handle these locations under react is


<details>
<summary>Eslint Error</summary>

```
Error: Cannot access refs during render

React refs are values that are not needed for rendering. Refs should only be accessed outside of render, such as in event handlers or effects. Accessing a ref value (the `current` property) during render can cause your component not to update as expected (https://react.dev/reference/react/useRef).

~/projects/forks/ocp/monitoring-plugin/web/src/components/dashboards/perses/hooks/useDashboardsData.ts:56:29
  54 |     const dashboardsChanged =
  55 |       persesDashboards.length !== prevDashboardsRef.current.length ||
> 56 |       persesDashboards.some((dashboard, i) => {
     |                             ^^^^^^^^^^^^^^^^^^^
> 57 |         const prevDashboard = prevDashboardsRef.current[i];
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 58 |         return (
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 59 |           dashboard?.metadata?.name !== prevDashboard?.metadata?.name ||
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 60 |           dashboard?.spec?.display?.name !== prevDashboard?.spec?.display?.name ||
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 61 |           dashboard?.metadata?.project !== prevDashboard?.metadata?.project
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 62 |         );
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 63 |       });
     | ^^^^^^^^ Passing a ref to a function may read its value during render
  64 |
  65 |     if (!dashboardsChanged && prevMetadataRef.current.length > 0) {
  66 |       return prevMetadataRef.current; (eslint react-hooks/refs)
```

</details>

web/package.json
jgbernalp: While testing this in detail, I found some typescript and lint issues. The `getCSRFToken` should not be used manually so I removed the usages in favor of the consoleFetch. PTAL: https://github.com/PeterYurkovich/monitoring-plugin/pull/2
I tested in a cluster and the POST request include the CSRF token correctly for dashboards creation, which was the code affected even it is not completely relevant for this PR.

PeterYurkovich: Thanks for putting the PR up, I didn't realize that the perses fetch function was only handling sessions. Since we are using the auth header on the backend now it shouldn't be necessary

PR #887, opened 2026-04-16
OBSINTA-1290: Incident Tests Stability

DavidRajnoha:
Testsuite stability improvement generated by POC agentic loop. Looking for coderabbit feedback.

- **OOM prevention in search loop**: `cy.reload()` to free browser DOM, `_quietSearch` to suppress                                                                                                                
    Cypress snapshot accumulation, and a 35-min hard timeout safety net against infinite hangs                                                                                                                     

- **Plugin warm-up**: `warmUpForPlugin()` polls for the Incidents tab up to 3 min, fixing flaky                                                                                                                   
    reg/02–05 specs caused by async plugin registration after session restoration                                                                                                                                   

- **Test hygiene**: remove `cy.pause()` debug calls, add missing fixture  

## Summary by CodeRabbit

* **Tests**
  * Improved incident test polling with explicit interval timing, OOM-safe polling, and a hard timeout to avoid infinite waits.
  * Added plugin/page warm-up steps and a reset for incident search timeout across suites for more reliable setup.
  * Removed manual pause points to speed runs and reduced noisy logging/snapshots.
  * Improved chart traversal/search reliability, added incident ID filter/traversal tests, and a new bar-selection walkthrough test.
  * Added a new incident scenario fixture for broader coverage.

* **Documentation**
  * Added a Test Stability Ledger summarizing per-test stability metrics.

### DIFF
--- a/web/cypress/views/incidents-page.ts
+++ b/web/cypress/views/incidents-page.ts
@@ -1,6 +1,18 @@
 import { nav } from './nav';
 import { DataTestIDs } from '../../src/components/data-test';
 
+// Hard timeout safety net for findIncidentWithAlert retry loops.
+// Prevents infinite loops if cy.waitUntil's timeout mechanism fails
+// (e.g., due to cy.reload() interfering with the Cypress command queue).
+let _findIncidentSearchStart: number | null = null;
+const _FIND_INCIDENT_HARD_TIMEOUT_MS = 35 * 60 * 1000; // 35 minutes
+
+// When true, search methods suppress Cypress command logging to prevent
+// DOM snapshot accumulation that causes OOM (exit 137) in CI containers.
+// Toggled by findIncidentWithAlert during its retry loop.
+let _quietSearch = false;
+const _qLog = (): { log: false } | Record<string, never> => (_quietSearch ? { log: false } : {});
+
 export const incidentsPage = {
   // Centralized element selectors - all selectors defined in one place
   elements: {
@@ -70,17 +82,17 @@ export const incidentsPage = {
     incidentsChartBar: (groupId: string) =>
       cy.byTestID(`${DataTestIDs.IncidentsChart.ChartBar}-${groupId}`),
     incidentsChartBarsVisiblePaths: () => {
-      return cy.get('body').then(($body) => {
+      return cy.get('body', _qLog()).then(($body) => {
         // There is a delay between the element being rendered and the paths being visible.
-        // The case when no paths are visible is valid, so we can not use should or conditional
-        // testing semantics.
-        cy.wait(500);
+        // The case when no paths are visible is valid, so we can not use should
+        // or conditional testing semantics.
+        cy.wait(500, _qLog());
         // We need to use the $body as both cases when the element is there or not are valid.
         const exists =
           $body.find('g[role="presentation"][data-test*="incidents-chart-bar-"]').length > 0;
         if (exists) {
           return cy
-            .get('g[role="presentation"][data-test*="incidents-chart-bar-"]')
+            .get('g[role="presentation"][data-test*="incidents-chart-bar-"]', _qLog())
             .find('path[role="presentation"]')
             .filter((index, element) => {
               const fillOpacity =
@@ -88,8 +100,8 @@ export const incidentsPage = {
               return parseFloat(fillOpacity || '0') > 0;
             });
         } else {
-          cy.log('Chart bars were not found. Test continues.');
-          return cy.wrap([]);
+          if (!_quietSearch) cy.log('Chart bars were not found. Test continues.');
+          return cy.wrap([], _qLog());
         }
       });
     },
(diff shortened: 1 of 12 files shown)


web/cypress/views/incidents-page.ts
```diff
@@ -70,26 +87,26 @@ export const incidentsPage = {
...
+        // The case when no paths are visible is valid, so we can not use should
+        // or conditional testing semantics.
+        cy.wait(500, _qLog());
         // We need to use the $body as both cases when the element is there or not are valid.
         const exists =
           $body.find('g[role="presentation"][data-test*="incidents-chart-bar-"]').length > 0;
         if (exists) {
           return cy
-            .get('g[role="presentation"][data-test*="incidents-chart-bar-"]')
+            .get('g[role="presentation"][data-test*="incidents-chart-bar-"]', _qLog())
```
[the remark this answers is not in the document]
DavidRajnoha: Valid observation — the `cy.wait(500)` inside the `.then()` callback is indeed queued after the synchronous `$body.find()` check, making the delay ineffective.

However, this is pre-existing code — our PR only added `_qLog()` to the existing calls. Additionally, `findIncidentWithAlert` no longer uses `incidentsChartBarsVisiblePaths()` (switched to `incidentsChartBarsGroups()` with a proper `$body.find()` guard), so this function is only used by chart hover/interaction helpers, not the search loop.

Fixing this here risks a regression in the hover helpers for no test coverage gain. Will track as a separate cleanup.

DavidRajnoha:
## Stability Fixes — Attribution & Impact Ranking

This PR contains fixes from both autonomous AI agent work and human-directed architectural decisions. Here's the breakdown ranked by impact:

### Stability Fixes (ranked by impact)

| Rank | Fix | Origin | Impact |
|---|---|---|---|
| **1** | **`cy.reload()` removal** | **Human-directed** | **Critical.** Eliminated an entire class of headless CI failures — `cy.reload()` caused the Console to re-resolve all dynamic plugin chunks from scratch, resulting in blank pages and silent test failures. Without this, tests couldn't run reliably in CI at all. |
| **2** | **Filter-based traversal** | **Human-directed** | **High.** Replaced inherently brittle chart bar clicks with deterministic filter dropdown interactions. This flattened Node.js memory (bar clicks generated ~40 Cypress commands per incident, filters generate ~5). Also eliminated multi-severity bar ambiguity and URL parameter conflicts. |
| **3** | **Plugin warm-up** (`warmUpForPlugin()` / `waitForIncidentsTab()`) | Agent-designed | **High.** Fixed the most frequent flaky failure — the Incidents tab not being registered when `beforeEach()` runs. Plugin registration after session restore takes 80-120s, but the default Cypress command timeout was 80s. The 3-minute jQuery-based polling eliminated this race entirely. |
| **4** | **Hard timeout safety net** (35 min kill switch) | Agent-designed | **Medium.** Prevents 2h+ CI hangs when `cy.waitUntil` fails to terminate. A backstop, not a primary fix — but critical for CI pipeline health since a hung job blocks the entire queue. |
| **5** | **Bar-group selection** (groups vs flattened paths) | Agent-designed | **Medium.** Fixed incorrect incident selection with multi-severity incidents where one bar produced multiple `<path>` elements. Partially superseded by #2 (filter-based traversal) for search flows, but still used in regression tests for chart interaction. |
| **6** | **OOM-safe BVT polling** (interval tuning) | Agent-designed | **Low-Medium.** Doubled the polling interval from 60s to 120s and wired in `findIncidentWithAlert` with its OOM protections. Reduced unnecessary churn during the 30-minute detection window. |
| **7** | **Quiet search** (`_quietSearch` / `_qLog()`) | Agent-designed | **Low.** Experiment confirmed: ~20% less Chrome heap growth, but Node.js stays flat regardless thanks to #2. A defense-in-depth measure with no downside, but not the fix that solved the OOM. |

### Summary

- **Human-directed** (2 fixes): `cy.reload()` removal, filter-based traversal — address **root causes**
- **Agent-designed** (5 fixes): quiet search, hard timeout, plugin warm-up, BVT polling, bar-group selection — address **symptoms and edge cases**

The top two fixes — both human-directed — target architectural issues (reload breaks plugins, bar clicks are inherently brittle). The agent correctly identified *symptoms* (OOM, hangs, race conditions) and applied *mitigations*, but the human identified the *root causes* and drove the *architectural* solutions.

Together they achieved **17/17 pass rate across 4 clusters and 8 runs**.

### Memory Experiment Results

A 6-arm experiment (3x quiet=ON, 3x quiet=OFF, 8 iterations each) confirmed:
- **Node.js RSS/heap**: completely flat at 167 MB / 65 MB across all arms — the filter-based traversal eliminated the original OOM risk
- **Chrome V8 heap**: quiet=ON grows ~136 MB vs quiet=OFF ~170 MB on average (**~20% reduction**)
- Experiment code preserved on branch `experiment/quiet-search-memory`

web/cypress/views/incidents-page.ts
```diff
@@ -1,6 +1,40 @@
...
+const _resetSearchState = () => {
+  _findIncidentSearchStart = null;
+  _quietSearch = false;
+};
+
+// jQuery selector for the Incidents tab — covers both PatternFly v6 and
+// legacy Console horizontal nav markup. Used by goTo()/warmUpForPlugin()
+// to poll for plugin registration without Cypress command overhead.
+const _INCIDENTS_TAB_SELECTOR =
+  '.pf-v6-c-tabs__item:contains("Incidents"), ' +
```
jgbernalp: We should avoid using patternfly specific selectors as this test will fail if we upgrade patternfly, prefer a test Id.

etmurasaki: yeah, I know sometimes it is difficult to add test ids due to console dependencies... but if at least, it could be encapsulated in the src/components/data-test.ts to centralize, even if adding the pf selector there... it would be good.

DavidRajnoha: I already changed that, only I could not verify as the job failed and none cluster I tried to create today succeeded.

PR #930, opened 2026-05-12
OU-1348: [release-coo-ocp-4.19]: fix: panel actions index so they are visible on click

jgbernalp:
Manual cherry-pick of https://github.com/openshift/monitoring-plugin/pull/928


jgbernalp:
/label qe-approved
to be tested in COO

PR #951, opened 2026-05-18
OCPBUGS-79456: immutable bump: fix for CVE-2026-29063 [4.20]

rhdmalone:
Fix CVE-2026-29063 (Prototype Pollution in the immutable npm package) by updating the dependency to patched versions.

Changes:

Bumped immutable in dependencies from 3.x to ^3.8.3 (the minimum patched version for the 3.x range)
Added overrides.sass.immutable: ^5.1.5 to force sass's transitive dependency on immutable to the patched 5.x version
Updated package-lock.json accordingly — both immutable@3.8.3 (direct) and immutable@5.1.5 (scoped under node_modules/sass) are now at their respective patched versions
CVE details:

CVE-2026-29063 — Prototype Pollution via mergeDeep(), mergeDeepWith(), merge(), Map.toJS(), Map.toObject()
Affected: < 3.8.3, >= 4.0.0-rc.1 < 4.3.8, >= 5.0.0-beta.1 < 5.1.5
Fixed: 3.8.3, 4.3.8, 5.1.5

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -67,7 +67,7 @@
         "i18next": "^21.8.14",
         "i18next-http-backend": "^2.2.0",
         "immer": "^10.1.3",
-        "immutable": "3.x",
+        "immutable": "^3.8.3",
         "js-yaml": "^4.1.0",
         "lodash-es": "^4.17.21",
         "lru-cache": "^6.0.0",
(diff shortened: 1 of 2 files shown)


rhdmalone:
Comments from Claude on e2e-monitoring failure on Tuesday May 19th
https://storage.googleapis.com/test-platform-results/pr-logs/pull/openshift_monitoring-plugin/951/pull-ci-openshift-monitoring-plugin-release-4.20-e2e-monitoring/2056666308753231872/build-log.txt

Comment 1:
"Now I can see the exact sequence. The test fails at line 56:

cy.byTestID(DataTestIDs.MetricGraphNoDatapointsFound).eq(0).scrollIntoView().should('be.visible');
The test:

Clicks the K8S_COMPUTE_RESOURCES_POD dashboard dropdown (line 55)
Then expects a [data-test="datapoints-msg"] "no datapoints" empty state element to appear (line 56)
The assumption is that the K8S_COMPUTE_RESOURCES_POD dashboard will have no data (showing the empty state), but in the CI environment it apparently does have data — so the datapoints-msg empty state element never renders.

This is a pre-existing flaky test — it relies on a specific dashboard having no datapoints in the CI cluster environment, which isn't guaranteed. It's nothing to do with your CVE changes."

Comment 2:

"There aren't actually 2 separate Cypress test failures — there is only 1 Cypress test failure (the datapoints-msg one). The 2 failing CI steps reported are:

e2e-monitoring-monitoring-plugin-tests-monitoring-ui — the container running the Cypress tests. It failed with exit code 1 because 1 Cypress test failed inside it.

Step phase test — this is the parent CI pipeline step that wraps the test container. It failed because step 1 failed — it's a cascading failure, not an independent one.

So in summary: one root cause (the datapoints-msg Cypress assertion in 03.reg_legacy_dashboards_admin.cy.ts), which causes two CI steps to be reported as failed. The second step failure is just the pipeline propagating the exit code upward."

PR #968, opened 2026-06-01
OU-1399: pass ISO string to Timestamp component for valid date rendering

tremes:
String(epochMs) produces a numeric string that new Date() cannot parse, resulting in Invalid Date and "-" displayed in the Start/End columns. Convert to ISO 8601 format via toISOString() instead.

Assisted-by: Claude Code:claude-opus-4-6

## Summary by CodeRabbit

* **Improvements**
  * Start and End timestamps now use standardized ISO formatting for consistent display.
  * When start or end times are unavailable (or an incident isn't resolved), the table shows '---' instead of a timestamp for clearer, more accurate incident status.


PeterYurkovich:
QE to be completed within COO

PR #976, opened 2026-06-08
CNV-87529: management: add alert rule classification system

sradco:
Add classification labels (component,
layer) for alert rules with validation,
ARC-based persistence, and bulk update
support.


docs/alert-rule-classification.md
```diff
@@ -0,0 +1,254 @@
...
+  - CA trust: system pool + `SSL_CERT_FILE` + `/var/run/configmaps/service-ca/service-ca.crt`.
+
+RBAC:
+- Read routes in `openshift-monitoring`.
+- Access `prometheuses/api` as needed for oauth-proxied endpoints.
+
+## Updating Rules Classification
+APIs:
+- Single update:
+  - Method: `PATCH /api/v1/alerting/rules/{ruleId}`
```
PeterYurkovich, line 191: open API spec is missing for the patch method, so `alert_rule_classification_patch.go` is never used. If the plan is to add it in a follow up PR then ignore

docs/alert-rule-classification.md
```diff
@@ -0,0 +1,254 @@
...
+Annotations are intentionally ignored to reduce id churn on documentation-only changes.
+
+## Classification Logic (How component/layer are determined)
+Location: `pkg/alertcomponent/matcher.go`
+
+1) The code adapts `cluster-health-analyzer` matchers:
+   - CVO-related alerts (update/upgrade) → component/layer based on known patterns
+   - Compute / node-related alerts
+   - Core control plane components (renamed to layer `cluster`)
+   - Workload/namespace-level alerts (renamed to layer `namespace`)
```
PeterYurkovich: Avoid using rename since alerts/rules have names:

```suggestion
   - Core control plane components (layer set to `cluster`)
   - Workload/namespace-level alerts (layer set to `namespace`)
```

pkg/classification/validation.go
```diff
@@ -0,0 +1,34 @@
...
+	"cluster":   {},
+	"namespace": {},
+}
+
+var labelValueRegexp = regexp.MustCompile(`^[A-Za-z0-9]([A-Za-z0-9_.-]*[A-Za-z0-9])?$`)
+var labelNameRegexp = regexp.MustCompile(`^[A-Za-z_][A-Za-z0-9_]*$`)
+
+// ValidateLayer returns true if the provided layer is one of the allowed values.
+func ValidateLayer(layer string) bool {
+	_, ok := allowedLayers[strings.ToLower(strings.TrimSpace(layer))]
```
PeterYurkovich: the allowed layers calls ToLower in a number of places, is there a reason that we should accept "Cluster" as a layer value instead of requiring "cluster"

pkg/management/update_classification.go
```diff
@@ -0,0 +1,459 @@
...
+		return &ValidationError{Message: "ruleId is required"}
+	}
+
+	if req.Component != nil && !classification.ValidateComponent(*req.Component) {
+		return &ValidationError{Message: fmt.Sprintf("invalid component %q", *req.Component)}
+	}
+	if req.Layer != nil && !classification.ValidateLayer(*req.Layer) {
+		return &ValidationError{Message: fmt.Sprintf("invalid layer %q (allowed: cluster, namespace)", *req.Layer)}
+	}
+	if req.ComponentFrom != nil {
```
PeterYurkovich, line 46: If `req.Component` is a whitespace only string then the validation error's. If `req.ComponentFrom` is a whitespace only string then the validation does NOT error. Is there a reason for the different handling of the 2 fields?

pkg/management/update_classification.go
```diff
@@ -0,0 +1,459 @@
...
+	existingOverrides, existingDrops := collectExistingFromARC(found, existingArc)
+	existingRuleDrops := getExistingRuleDrops(existingArc, alertRuleId)
+	effective := computeEffectiveLabels(original, existingOverrides, existingDrops)
+
+	desired := buildDesiredLabels(effective, classificationLabels)
+	nextChanges := buildNextLabelChanges(original, desired)
+
+	if len(nextChanges) == 0 {
+		if found {
+			if err := c.k8sClient.AlertRelabelConfigs().Delete(ctx, arcNamespace, arcName); err != nil {
```
PeterYurkovich, line 196: Since the `len(nextChanges)` is called without regard to existingRuleDrops, this code will always attempt to delete an ARC if there are no changes after applying the ARC, which may not be desired. An ARC that was manually created with a drop command on a specific label which does not currently exist on the `PrometheusRule` will be deleted even if the user wanted the ARC to exist to gate against that label being added in the future

pkg/management/update_classification.go
```diff
@@ -0,0 +1,459 @@
...
+		AdditionalInfo: fmt.Sprintf("in PrometheusRule %s/%s", namespace, name),
+	}
+}
+
+// ApplyDynamicClassification resolves the effective component and layer for an
+// alert by applying _from indirection. If a rule carries a component_from or
+// layer_from label, the corresponding alert label value is used instead of the
+// static default. Unresolvable or empty lookups fall back to the supplied
+// defaults.
+func ApplyDynamicClassification(ruleLabels, alertLabels map[string]string, defaultComponent, defaultLayer string) (string, string) {
```
PeterYurkovich, line 447: I assume this will be used in a future PR for get AlertRules, just want to double check

docs/alert-rule-classification.md
```diff
@@ -0,0 +1,254 @@
...
+- `labels`: only non-system labels
+  - excludes labels with `openshift_io_` prefix and the `alertname` label
+  - drops empty values
+  - keeps only valid Prometheus label names (`[a-zA-Z_][a-zA-Z0-9_]*`)
+  - sorted by key and joined as `key=value` lines
+
+Annotations are intentionally ignored to reduce id churn on documentation-only changes.
+
+## Classification Logic (How component/layer are determined)
+Location: `pkg/alertcomponent/matcher.go`
```
PeterYurkovich, line 44: This code isn't included in this PR, which is fine, but I'm wondering how this will be implemented once it is. Will there be some kind of reconcile loop which watches for PrometheusRules, or will it only trigger an update when one of these API endpoints are called?

pkg/management/update_classification.go
```diff
@@ -0,0 +1,459 @@
...
+		if arc.Labels == nil {
+			arc.Labels = map[string]string{}
+		}
+		arc.Labels[managementlabels.ARCLabelPrometheusRuleNameKey] = prName
+		arc.Labels[managementlabels.ARCLabelAlertNameKey] = alertName
+		if arc.Annotations == nil {
+			arc.Annotations = map[string]string{}
+		}
+		arc.Annotations[managementlabels.ARCAnnotationAlertRuleIDKey] = alertRuleId
+		if err := k8sClient.AlertRelabelConfigs().Update(ctx, *arc); err != nil {
```
PeterYurkovich, line 392: Not needed for this PR, but is there any desire to have retry functionality occur when trying to update, or should that be something that is strictly driven by the frontend UI (ie. user needs to resubmit a form)

docs/alert-rule-classification.md
```diff
@@ -0,0 +1,254 @@
...
+2. Apply each classification label as a `Replace` action keyed on `openshift_io_alert_rule_id`.
+
+When all overrides are removed, the ARC is deleted.
+
+**AlertingRule CR distinction:** Some platform alerts are defined via `AlertingRule` CRs,
+which the cluster-monitoring-operator reconciles into `PrometheusRule` resources. When
+the owning `AlertingRule` CR is operator-managed (has operator owner references), the
+backend cannot modify it directly (the operator would reconcile the change back). In
+this case, label updates are applied through an ARC instead. When the `AlertingRule` CR
+is not externally managed, label updates are written directly into the CR. Classification
```
PeterYurkovich: I'm unclear on why we would want to swap between directly updating the `AlertingRule` CR and creating/updating an ARC through a single backend endpoint instead of always just always updating through the ARC when this endpoint is called.

I can image a scenario where the frontend has created a user-created and managed AR which has these component and layer labels, and then the user wants to update these labels. It is nice to have these validation layers for these labels specifically, but if a user manually updates the labels for the alert rather than using this endpoint then the validation will be lost.

It seems like it would make more sense to have this "update the ARC vs update the AR" decision happen on the frontend rather than the backend, and then have the update AR logic always perform the validation on these labels if they exist. That way the PATCH requests with the `classification` will only ever update ARC's and an update to an AR directly will still receive the validation

docs/alert-rule-classification.md
sradco: Yes, the OpenAPI spec for the PATCH endpoints will be added in a follow-up PR. This PR focuses on the backend classification logic and ARC persistence; the API wiring (OpenAPI spec + handler in `alert_rule_classification_patch.go`) will come in the next PR in the stack.

docs/alert-rule-classification.md
sradco: Good catch, I applied the suggestion. Updated to "layer set to" to avoid confusion with alert/rule names.

pkg/classification/validation.go
sradco: +1. I removed the `ToLower`/`TrimSpace` from `ValidateLayer` ,  it now requires exact lowercase (`cluster` or `namespace`). 
The normalization in `buildClassificationLabels` was also removed since validation already ensures correctness.
This makes the API strict at the boundary.

pkg/management/update_classification.go
sradco: No good reason for the difference, fixed. 
`ComponentFrom` and `LayerFrom` now reject whitespace-only values with a clear error message ("must not be empty or whitespace-only; set to null to remove"), consistent with how `Component` rejects empty strings via `ValidateComponent`.

pkg/management/update_classification.go
sradco: Good catch. Fixed. The ARC deletion check now also considers `existingRuleDrops`. 
The condition is `len(nextChanges) == 0 && len(existingRuleDrops) == 0` before deleting. 
If there are existing Drop configs on the ARC (e.g., a disabled rule), the ARC is preserved even when classification produces no label changes.

pkg/management/update_classification.go
sradco: Correct. `ApplyDynamicClassification` is used in the get-alerts read path (`get_alerts.go`) to resolve `_from` labels at query time. 
It's called from the alerts enrichment flow when building the API response. 
The get-alerts PR is the next one in the stack.

docs/alert-rule-classification.md
sradco: No reconcile loop. Classification is triggered only through the API endpoints (PATCH single/bulk). 
The matcher (`pkg/alertcomponent/matcher.go`) runs at read time during alert enrichment. 
When `GET /api/v1/alerting/alerts` is called, the backend computes component/layer from rule labels and the matcher fallback logic, then applies any ARC-based overrides. 
The matcher code itself is included in the base branch (`main-alerts-management-api`) and was introduced in an earlier PR in the stack.

pkg/management/update_classification.go
sradco: Good question. For now, retry is driven by the frontend/client. 
If an ARC update fails (e.g., conflict on `resourceVersion`), the client should re-fetch and resubmit. Backend retry with conflict resolution is worth considering for bulk operations at scale, but for MVP the client-driven approach is simpler and matches how other management endpoints (severity, drop/restore) work today.

docs/alert-rule-classification.md
sradco: Good feedback. I've clarified the doc.
The classification endpoint in this PR **always uses the ARC path**, regardless of whether the AlertingRule CR is operator-managed or not. It never writes directly to AlertingRule CRs.

The paragraph you flagged was describing the behavior of *other* management endpoints (severity, drop/restore) which may write to unmanaged AlertingRule CRs directly. I've rewritten it to make this distinction clear:

> Classification overrides always use the ARC path regardless of AlertingRule management status — this endpoint never writes directly to AlertingRule CRs. (Other management endpoints, such as severity updates, may write to unmanaged AlertingRule CRs directly, but classification is ARC-only.)

This is intentional because classification labels are additive metadata that should not interfere with the rule definition itself, and the ARC path provides a consistent storage model for both platform and user-defined rules.

PeterYurkovich:
/hold 
hold for Simon's review

sradco:
@simonpasquier Please review this PR. Thank you.

sradco:
Hi @simonpasquier, Can you please review this PR?

PR #981, opened 2026-06-09
OBSDA-609: feat: add multi-namespace selector to Developer Metrics page

prithvipatil97:
## Summary

Non-admin users in the Developer Console (Observe > Metrics) can currently only query metrics for one namespace at a time. This PR adds a multi-namespace selector dropdown that allows querying across multiple namespaces simultaneously, without requiring cluster-admin privileges.

- Adds a PatternFly v6 multi-select namespace dropdown to the Metrics page (Developer perspective only)
- Uses the OpenShift Project API (`project.openshift.io`) to list only namespaces the user has RBAC access to
- Extends `PrometheusURLProps.namespace` to accept `string | string[]`, appending each as a separate `?namespace=` query parameter
- Includes "All Namespaces" toggle and a 25-namespace safeguard cap to prevent Thanos overload

### Key finding

The backend monitoring stack (thanos-querier, prom-label-proxy, kube-rbac-proxy, console proxy) **already supports multi-namespace queries natively**. This change is **frontend-only** — no backend, operator, or cluster-level modifications are required.

### Files changed

**New files:**
- `web/src/components/hooks/useMultiNamespace.ts` — Hook managing multi-namespace selection state via ProjectModel
- `web/src/components/multi-namespace-selector.tsx` — PatternFly v6 multi-select dropdown component

**Modified files:**
- `web/src/components/utils.ts` — Extended namespace type and URL builder for arrays
- `web/src/components/MetricsPage.tsx` — Integrated selector and threaded `namespaceOverride` to downstream components
- `web/src/components/query-browser.tsx` — Added optional `namespace` prop to `QueryBrowserProps`

### Design decisions

- **Project API instead of Namespace API** — non-admin users cannot list cluster-scoped Namespaces; the Project API returns only projects the user has access to (pattern already used in `useOcpProjects`, `useActiveProject`)
- **Developer perspective only** — selector renders only when `useMetricsTenancy` is true; Admin perspective already has cluster-wide access
- **Backward compatible** — single namespace selection behaves identically to current behavior

## Screenshots

**Before** — single project selector, one namespace at a time:

Before

**After** — multi-namespace selector with all namespaces selected and cross-namespace graph:

After

## Test plan

- [x] Verified backend supports multi-namespace queries via `oc port-forward` to thanos-querier:9092
- [x] Built and deployed custom image to OCP 4.22 cluster
- [x] Logged in as non-admin user (`titan`) with access to app1, app2, app3
- [x] Multi-namespace selector appears in Developer > Observe > Metrics
- [x] Cross-namespace metric queries return correct results
- [x] "All Namespaces" toggle selects all accessible projects
- [x] RBAC boundaries maintained — no unauthorized data exposed
- [x] Single namespace selection works identically to current behavior
- [x] Admin perspective unaffected — selector does not appear

Jira: https://redhat.atlassian.net/browse/OBSDA-609

## Summary by CodeRabbit

* **New Features**
  * Multi-namespace selector for metrics queries with support for up to 25 namespaces.
  * Select all / deselect all and toggle behavior with visual selection indicators and count badge.
  * Per-query execution now supports single or multiple namespaces.
  * Query building and polling respect the selected namespace(s), including handling of the “all namespaces” case.

### DIFF
--- a/web/src/components/MetricsPage.tsx
+++ b/web/src/components/MetricsPage.tsx
@@ -118,6 +118,8 @@ import { MonitoringProvider } from '../contexts/MonitoringContext';
 import { DataTestIDs } from './data-test';
 import { useMonitoring } from '../hooks/useMonitoring';
 import { useMonitoringNamespace } from './hooks/useMonitoringNamespace';
+import { useMultiNamespace } from './hooks/useMultiNamespace';
+import { MultiNamespaceSelector } from './multi-namespace-selector';
 import { useSearchParams } from 'react-router';
 
 // Stores information about the currently focused query input
@@ -952,7 +954,8 @@ const Query: FC<{
   index: number;
   customDatasource?: CustomDataSource;
   units: GraphUnits;
-}> = ({ index, customDatasource, units }) => {
+  namespaceOverride?: string | string[];
+}> = ({ index, customDatasource, units, namespaceOverride }) => {
   const { t } = useTranslation(process.env.I18N_NAMESPACE);
   const { plugin } = useMonitoring();
 
@@ -1069,7 +1072,7 @@ const Query: FC<{
         <QueryTable
           index={index}
           customDatasource={customDatasource}
-          namespace={activeNamespace}
+          namespace={namespaceOverride ?? activeNamespace}
           units={units}
         />
       </DataListContent>
@@ -1082,7 +1085,14 @@ const QueryBrowserWrapper: FC<{
   customDataSource: CustomDataSource;
   customDatasourceError: boolean;
   units: GraphUnits;
-}> = ({ customDataSourceName, customDataSource, customDatasourceError, units }) => {
+  namespaceOverride?: string | string[];
+}> = ({
+  customDataSourceName,
+  customDataSource,
+  customDatasourceError,
+  units,
+  namespaceOverride,
+}) => {
   const { t } = useTranslation(process.env.I18N_NAMESPACE);
   const { plugin } = useMonitoring();
   const [activeNamespace] = useActiveNamespace();
@@ -1209,6 +1219,7 @@ const QueryBrowserWrapper: FC<{
     <QueryBrowser
       customDataSource={customDataSource}
       disabledSeries={disabledSeries}
+      namespace={namespaceOverride}
       queries={queryStrings}
       units={units}
       showStackedControl
@@ -1263,10 +1274,11 @@ const RunQueriesButton: FC = () => {
   );
 };
 
-const QueriesList: FC<{ customDatasource?: CustomDataSource; units: GraphUnits }> = ({
-  customDatasource,
-  units,
-}) => {
+const QueriesList: FC<{
+  customDatasource?: CustomDataSource;
+  units: GraphUnits;
+  namespaceOverride?: string | string[];
+}> = ({ customDatasource, units, namespaceOverride }) => {
   const { plugin } = useMonitoring();
   const count = useSelector(
     (state: MonitoringState) => getObserveState(plugin, state).queryBrowser.queries.length,
@@ -1282,6 +1294,7 @@ const QueriesList: FC<{ customDatasource?: CustomDataSource; units: GraphUnits }
             key={reversedIndex}
             customDatasource={customDatasource}
             units={units}
+            namespaceOverride={namespaceOverride}
           />
         );
       })}
@@ -1339,7 +1352,8 @@ const MetricsPage_: FC = () => {
   const [units, setUnits] = useQueryParam(QueryParams.Units, StringParam);
   const [customDataSourceName] = useQueryParam(QueryParams.Datasource, StringParam);
   const { namespace, setNamespace } = useMonitoringNamespace();
-  const { displayNamespaceSelector } = useMonitoring();
+  const { displayNamespaceSelector, useMetricsTenancy } = useMonitoring();
+  const multiNs = useMultiNamespace();
 
   const dispatch = useDispatch();
 
@@ -1456,6 +1470,18 @@ const MetricsPage_: FC = () => {
       </ListPageHeader>
       <PageSection hasBodyWrapper={false}>
         <Stack hasGutter>
+          {useMetricsTenancy && (
+            <StackItem>
+              <MultiNamespaceSelector
+                accessibleNamespaces={multiNs.accessibleNamespaces}
+                selectedNamespaces={multiNs.selectedNamespaces}
+                allSelected={multiNs.allSelected}
+                toggleNamespace={multiNs.toggleNamespace}
+                toggleAll={multiNs.toggleAll}
+                deselectAll={multiNs.deselectAll}
+              />
+            </StackItem>
+          )}
           <StackItem>
             <ToggleGraph />
           </StackItem>
@@ -1465,6 +1491,7 @@ const MetricsPage_: FC = () => {
               customDataSourceName={customDataSourceName}
               customDatasourceError={customDatasourceError}
               units={units as GraphUnits}
+              namespaceOverride={useMetricsTenancy ? multiNs.namespaceForQuery : undefined}
             />
           </StackItem>
           <StackItem>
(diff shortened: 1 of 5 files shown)


PeterYurkovich:
I think it doesn't make sense to have this feature be present in the developer perspective which is focused on single namespace views. But it could make sense in the admin perspective (which non-admin users have access to use now). That being said, we should either integrate or replace the namespace bar on the metrics page with this new selector. Maybe we can get some input from @fkargbo on what that should look like. My guess would be fully removing the namespace bar from the top and then having a namespace selector elsewhere similar to what you have

prithvipatil97:
Thanks for your attention here and for the review, @PeterYurkovich. The UX feedback around replacing the namespace bar rather than adding a separate selector makes sense.

Sure, we will wait for the input from @fkargbo.

Regarding moving this to the Admin perspective instead of Developer — I wanted to flag a technical consideration:

The Developer perspective uses the tenancy-based Thanos endpoint (port 9092), which routes through prom-label-proxy and kube-rbac-proxy to enforce per-namespace RBAC. This is what allows non-admin users to query metrics scoped to their own namespaces — our implementation sends multiple ?namespace= parameters to this endpoint, and each is individually authorized.

The Admin perspective, on the other hand, uses the non-tenancy endpoint (port 9091), which typically requires the cluster-monitoring-view ClusterRole. Non-admin users generally don't have this permission, so moving the feature there could mean they wouldn't be able to query metrics at all — which is the exact limitation [OBSDA-609](https://redhat.atlassian.net/browse/OBSDA-609) is trying to solve.

That said, if there's been recent work to wire the Admin perspective metrics page to also support tenancy-based queries for non-admin users, then the Admin perspective could work. Could you clarify if that's the case?

Regards,
Prithviraj Patil

PeterYurkovich:
> That said, if there's been recent work to wire the Admin perspective metrics page to also support tenancy-based queries for non-admin users, then the Admin perspective could work. Could you clarify if that's the case?

That would be correct. In the admin perspective metrics page there is now a namespace dropdown at the top of the page (similar to how the dev perspective looks), which uses the non-tenancy path when "All Namespaces" is selected and the tenancy path when any other namespace is selected

fkargbo:
Hi, @PeterYurkovich @prithvipatil97 the images didn't work for me. It would be good to see what is proposed for context from a UX perspective. You can also send me screenshots on Slack. Thanks!

PeterYurkovich:
Images are available in the Jira @fkargbo

prithvipatil97:
Hello @PeterYurkovich , 
Thanks for clarification about Admin perspective tenancy support.
Sure,  I will wait for UX direction.

Hello @fkargbo , I have also attached the screenshots in the Jira [OBSDA-609](https://redhat.atlassian.net/browse/OBSDA-609).
Let me also share the screenshots over a slack.

Regards,
Prithviraj Patil

fkargbo:
Hi Prithvi, it looks good to me over all. My understanding is that "Project" and "Namespace" mean the same thing. I'm wondering will there be a conflict in selecting a Project at the top then selecting a "namespace" on the second selector?

prithvipatil97:
Hello @fkargbo 

I think you're right. Currently, the top-level Project selector sets the active namespace, while our multi-namespace selector overrides it for metric queries. This can be confusing when the two show different selections.

As discussed on Slack, we have two possible approaches:
- Completely remove the top-level namespace selector from the Metrics page and replace it with a multi-namespace selector.
- Enhance the existing namespace dropdown to support multi-select functionality.

You recommended the second option, and @PeterYurkovich  also suggested that we should extend the existing namespace selector rather than introducing a separate one.

I will proceed with this approach, do some more investigation, and come back with my findings.

Regards,
Prithviraj Patil

PeterYurkovich:
@prithvipatil97 Sorry to yank you back and forth, but I think it would probably be best to go with option 1. The namespace selector is a shared component which is used across most pages of the OpenShift console. Even if we implement a change which allows it to select multiple namespaces, users will be conditioned to expect it to only work for a single one. I don't believe that changing the namespace bar for this feature would be the best idea due to user expectations

prithvipatil97:
Hello @PeterYurkovich ,
Okay. Thanks for the suggestion. Sure, let me work on option 1 as well.

Currently, as per yesterday's discussion, I investigated more on option 2, and this is how the new design looks [Enhance the existing namespace dropdown to support multi-select functionality.]

[OBSDA-609](https://redhat.atlassian.net/browse/OBSDA-609)_New_working_fix

Requesting both of you to review this new design. I will continue exploring option 1 in parallel.


Regards,
Prithviraj Patil

PeterYurkovich:
The namespace bar should retain some of it's original features (specifically favorites and searching). Users will have access to hundreds of namespaces in large clusters and need ways to select their commonly used ones and filter to others quickly

fkargbo:
@prithvipatil97 I echo what @PeterYurkovich said. We can reuse the same dropdown component and adjust it to be multi-select capable. Also, are we changing from "Projects" to "Namespace(s) in the UI for this component?

prithvipatil97:
Hello @PeterYurkovich, @fkargbo ,
Thanks for your response.

Okay, that makes sense, and this will be a useful addition, especially with the original features like favorites and search.

The current NamespaceBar / namespace dropdown is a shared component from the `console SDK` (@openshift-console/dynamic-plugin-sdk), which lives in the `openshift/console` repo -- not in `monitoring-plugin`.

I apologize for the repeated questions, but could you please confirm how we should proceed?

Should we:

A) Build a custom multi-select project dropdown within the `monitoring-plugin` repo that replicates the existing features (favorites, search/filtering), similar to how the Perses dashboards already have a custom `ProjectDropdown` with these features? This would only affect the Metrics page.

B) Contribute to the `openshift/console` repo to make the shared NamespaceBar component natively support multi-select? This would be a broader scope but would benefit all console pages.

@fkargbo , Regarding your last query: ` are we changing from "Projects" to "Namespace(s) in the UI for this component?` 
==> No, I was just testing it. I will keep "Projects" to stay consistent with the rest of the console UI. 

Regards,
Prithviraj Patil

PeterYurkovich:
> We can reuse the same dropdown component and adjust it to be multi-select capable
We can't "re-use" because the implementation is located in the `openshift/console` codebase and it is very complicated (implementation is hundreds of lines across many files). The dropdown is adjusted frequently for each release and is a core part of the console and how it is structured, so changes to the `openshift/console` codebase would require changes in a large number of pages to support it. We have done something similar to cloning the dropdown in the past with the perses project selector, and it is very much a moving target and is difficult to maintain

> Also, are we changing from "Projects" to "Namespace(s) in the UI for this component?
You are correct we should use "Projects" instead to be consistent with the console 
image

Perhaps it would be best to schedule a call to discuss the UX and design of this task so that we don't have Prithviraj going back and forth so much

fkargbo:
@PeterYurkovich sounds good to me

prithvipatil97:
Hello @PeterYurkovich, @fkargbo,

Apologies for the delayed response, I was engaged with the Logging 6.6 hackathon and a few other priorities over the past few weeks.

I agree that a call would be the most efficient way to align on the UX and design for this feature. That way we can finalize the approach and avoid further back-and-forth.

I will soon share a meeting link with you.

Regards,
Prithviraj Patil

PeterYurkovich:
@prithvipatil97 following up on this, would you still like to have a call to discuss this?

prithvipatil97:
Hello @PeterYurkovich ,
Yes.

I was checking your calendar.
Can we schedule a meeting on Monday, 24th August 2026 at 10:00 AM EST? [Which is 07:30 PM IST (Indian Standard Time) for me].

Kindly confirm?

Attaching the screenshot on my latest work:
Monitoring_plugin-OBSDA-609


Regards
Prithviraj Patil

PeterYurkovich:
Hi @prithvipatil97 I'd be happy to meet, but I feet it relevant to say that our team is going to try and prioritize integrating the perses explore views into the console sooner than we were originally planning to do so. These explore view will contain a superset of the metrics page, and will allow users substantially more flexible data viewing such as switching between datasources (in cluster monitoring vs MonitoringStack deployed via COO) and other benefits. . If you would like to check it out you can check out the demo for perses here: https://demo.perses.dev/

I will keep my time open for the slot you suggested if you still feel it is necessary, but if you feel that this update will resolve your issue please let me know

image

prithvipatil97:
Hello @PeterYurkovich ,
Good Morning.
Thanks for your response and for sharing the new plan regarding the perses explore views.

Okay. Let me take a look at it and check out the demo for perses.

Then let's not connect today. I will keep you posted here.

Have a nice day.

regards,
Prithviraj Patil.

PR #985, opened 2026-06-10
OBSINTA-1387: add COO-LightSpeed show_timeseries e2e test

tremes:
Add e2e test for the OLS show_timeseries tool integration. Tagged with @ols to exclude from common e2e runs — requires both COO and OLS operators installed on the cluster.

Assisted-by: Claude Code:claude-opus-4-6


etmurasaki:
in a future PR we can enhance this automation to get setup of COO installation and OLS config done. 
good job

tremes:
@etmurasaki I am not sure if we want to have the COO & OLS installation here, but open to your suggestions. We definitely need to extend it to cover the "add to dashboard" functionality.

PR #995, opened 2026-06-11
OU-1384: remove global eslint disables

PeterYurkovich:
## Release Notes

* **Tests**
  * Enhanced testing infrastructure with new mock components for improved test coverage.

* **Chores**
  * Improved code quality by updating linting configuration and best practices.
  * Enhanced component metadata for better debugging and development tooling.
  * Reformatted configuration files for consistency and maintainability.

### DIFF
--- a/web/cypress.config.ts
+++ b/web/cypress.config.ts
@@ -25,16 +25,18 @@ export default defineConfig({
   },
   env: {
     grepFilterSpecs: true,
-    HOST_API: (process.env.CYPRESS_BASE_URL || '').replace(/console-openshift-console.apps/, 'api').concat(
-      ':6443',
-    ),
+    HOST_API: (process.env.CYPRESS_BASE_URL || '')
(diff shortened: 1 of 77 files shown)


PeterYurkovich:
Ignoring coderabbit review on purpose. This PR isn't to fix all of the lint issues, just move them to their own files rather than as global overrides

jgbernalp:
I think it will be better to actually resolve some issues rather than adding an ignore rule, for example the prop types and any assertions. We probably can split this PR into individual ones that disable one rule and fix the issues. WDYT?

I'm not sure I understand the value of moving the rules from global to scoped files if we are still ignoring them.

PeterYurkovich:
The benefit is it enables lint rules on all new code written and allows for follow up PR's to clean up the individual overrides. It's not a huge deal either way, but this way does enable the lint rules sooner

PR #1004, opened 2026-06-15
OU-1384: Add missing prop types

PeterYurkovich:
## Summary by CodeRabbit

* **Refactor**
  * Improved TypeScript type safety across React components by adding explicit prop typings (including `memo`-based components) and removing prop-type lint suppressions.
  * Updated Cypress mock components to use typed props and more consistent prop passing for stable test rendering.
  * Removed an unused `VirtualizedTable` mock export from the Cypress dummy setup.
  * Made the `health` prop optional in the targets page UI component (typing only).
  * Exported shared prop types for OLS tool UI components to enable reuse.


PeterYurkovich:
Overriding long running flakey test so I can continue with the lint rule updates

PR #1010, opened 2026-06-23
OCPBUGS-85473: fix: escape regex variable values in regex matcher contexts

jgbernalp:
This PR moves the `evaluateVariableTemplate` into its own utility file and adds tests for the regex escape bug.

## Summary by CodeRabbit

* **Refactor**
  * Centralized dashboard variable-template evaluation into a shared utility, including support for interval/range-derived placeholders and consistent namespace handling.
  * Improved substitution behavior for “all” selections and PromQL regex contexts to ensure generated queries remain valid.

* **Tests**
  * Added a Jest suite covering variable substitution, escaping rules for different matcher types, IPv6/regex escaping, namespace overrides, loading/undefined cases, and combined template scenarios.

### DIFF
--- a/web/src/components/dashboards/legacy/variable-utils.ts
+++ b/web/src/components/dashboards/legacy/variable-utils.ts
@@ -1,0 +1,88 @@
+import * as _ from 'lodash-es';
+import { DEFAULT_GRAPH_SAMPLES, MONITORING_DASHBOARDS_VARIABLE_ALL_OPTION_KEY } from './utils';
+import { ALL_NAMESPACES_KEY } from '../../utils';
+
+const intervalVariableRegExps = ['__interval', '__rate_interval', '__auto_interval_[a-z]+'];
+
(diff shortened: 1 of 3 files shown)


web/src/components/dashboards/legacy/variable-utils.ts
```diff
@@ -0,0 +1,94 @@
+import * as _ from 'lodash-es';
+import { DEFAULT_GRAPH_SAMPLES, MONITORING_DASHBOARDS_VARIABLE_ALL_OPTION_KEY } from './utils';
+
+const ALL_NAMESPACES_KEY = '#ALL_NS#';
```
PeterYurkovich: ALL_NAMESPACES_KEY should be imported

jgbernalp:
/cherry-pick release-4.22 release-4.21 release-4.20 release-4.19

PR #1012, opened 2026-06-24
NO-JIRA: prefer KUBECONFIG over in-cluster config for k8s client

sradco:
## Summary
- Fix e2e-management-api CI test failures by addressing four issues discovered during CI debugging:
  1. **Wrong cluster**: Server used `rest.InClusterConfig()` which connected to the build cluster instead of the provisioned test cluster. Introduce `loadKubeConfig()` that prefers KUBECONFIG when set.
  2. **Dead informers**: The k8s client was created with a 30-second timeout context. When `createHTTPServer` returned, `defer initCancel()` cancelled the context and stopped all informers — freezing the relabeled-rules cache permanently. Pass the server-lifetime context to `NewClient` instead.
  3. **Missing auth token**: The e2e framework only read `config.BearerToken`, which is empty when the CI kubeconfig uses client-certificate auth. Add `resolveToken()` that falls back to creating a ServiceAccount token via the TokenRequest API.
  4. **Spec-equivalence conflicts**: All test rules used `vector(1)`, triggering 409 conflicts across tests. Give each rule a unique PromQL expression.

## Root Cause
The CI step `monitoring-plugin-tests-management-api` runs on the build cluster (build01) with KUBECONFIG pointing to the provisioned test cluster. Multiple issues prevented the e2e tests from working:
- The server used `rest.InClusterConfig()` connecting to the wrong cluster
- Informers were killed after init due to a cancelled timeout context
- No bearer token was available for HTTP requests (client-cert kubeconfig)
- Shared PromQL expressions caused duplicate-spec rejections

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./pkg/... ./internal/...` passes
- [x] `go vet ./...` passes
- [ ] CI e2e-management-api test passes

### DIFF
--- a/test/e2e/delete_alert_rule_test.go
+++ b/test/e2e/delete_alert_rule_test.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"testing"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -41,6 +42,10 @@ func TestDeleteAlertRule(t *testing.T) {
 	}
 
 	t.Logf("Created 3 rules with IDs: %v", ruleIDs)
+
+	// Allow time for the informer watch event to propagate and
+	// the relabeled-rules cache to sync the new PrometheusRule.
+	time.Sleep(2 * time.Second)
 
 	deleteReq := managementrouter.BulkDeleteAlertRulesRequest{
 		RuleIds: []string{ruleIDs[0], ruleIDs[1]},
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -12,9 +12,13 @@ import (
 
 	osmv1client "github.com/openshift/client-go/monitoring/clientset/versioned"
 	monitoringv1client "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+	authv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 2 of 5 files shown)


test/e2e/framework/framework.go
```diff
@@ -73,12 +73,21 @@ func New() (*Framework, error) {
 		return nil, fmt.Errorf("failed to create osmv1 clientset: %w", err)
 	}
 
+	bearerToken := config.BearerToken
+	if bearerToken == "" && config.BearerTokenFile != "" {
+		tokenBytes, err := os.ReadFile(config.BearerTokenFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read bearer token file %q: %w", config.BearerTokenFile, err)
```
simonpasquier: err should already include the path
```suggestion
			return nil, fmt.Errorf("failed to read bearer token file: %w", err)
```

simonpasquier:
@coderabbitai review

PeterYurkovich:
/retitle NO-JIRA: prefer KUBECONFIG over in-cluster config for k8s client

test/e2e/delete_alert_rule_test.go
```diff
@@ -42,6 +43,10 @@ func TestDeleteAlertRule(t *testing.T) {
 
 	t.Logf("Created 3 rules with IDs: %v", ruleIDs)
 
+	// Allow time for the informer watch event to propagate and
+	// the relabeled-rules cache to sync the new PrometheusRule.
+	time.Sleep(2 * time.Second)
```
simonpasquier, line 48: this is going to be flaky: a better option is to
1. create the rules
2. check that they are present
3. delete them

All 3 steps should use a retry mechanism in case the API has a blip.

PR #1014, opened 2026-06-25
OU-1409: feat: fetch lazily ai proposal for each row in the alert list

jgbernalp:
https://github.com/user-attachments/assets/2455ba82-8f62-47e1-a577-2ee925e447fd

## Summary by CodeRabbit

* **New Features**
  * Added AI investigation support in alert views, including a new action to open related investigations when available.
  * Alerts can now surface investigation loading state while matching results are fetched.
  * Updated alert-related labels and messages for clearer UI wording.

* **Bug Fixes**
  * Improved alert-to-investigation matching for more reliable results.
  * Added support for viewing investigations from alert actions across the alerts page.


web/src/components/hooks/usePerspective.tsx
```diff
@@ -318,3 +318,20 @@ export const getDashboardsListUrl = (perspective: Perspective) => {
...
+  switch (perspective) {
+    case 'acm':
+      return `/multicloud/monitoring/v2/ai/proposals${queryParams}`;
+    case 'dev':
+      return `/dev-monitoring/v2/ai/proposals${queryParams}`;
+    case 'virtualization-perspective':
+      return `/virt-monitoring/v2/ai/proposals${queryParams}`;
+    case 'admin':
+    default:
+      return `/monitoring/v2/ai/proposals${queryParams}`;
```
PeterYurkovich, line 328: Generally we are returning empty string for perspectives we don't support, I think we should do the same here. I think this will only be in the admin perspective

PR #1028, opened 2026-07-03
Fix for OCPBUGS-94005: CVE-2026-13676

germanparente:
## Summary
- Fixes CVE-2026-13676: security policy bypass via improper Unicode hostname canonicalization in `fast-uri` (CVSS 7.5)
- Adds `fast-uri` override (`>=3.1.3`) in `web/package.json`, resolving to 4.1.0
- Follows established CVE fix pattern (npm overrides)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


web/package.json
```diff
@@ -162,7 +162,7 @@
   },
   "overrides": {
     "echarts": "^5.6.0",
-    "fast-uri": "^3.1.2",
+    "fast-uri": ">=3.1.3",
```
jgbernalp: This makes the version change to the next major version (4.1.0) which might be incompatible. There is a patched version for the current major version being used (3.1.3). Please adjust so the right major version is targeted in this update.

germanparente: thanks.

germanparente:
As there's a problem  with automation:
"expected dependent [Jira Issue OCPBUGS-94056](https://redhat.atlassian.net/browse/OCPBUGS-94056) to be in one of the following states: MODIFIED, ON_QA, VERIFIED, but it is Closed (Done) instead"

I will skip jira dependency:

/label jira/skip-dependent-bug-check

PR #1034, opened 2026-07-06
OU-1389: swap monitoring-plugin to be feature driven

PeterYurkovich:
This PR is part of a set that looks to update the monitoring-plugin to be feature driven, allowing for each feature to be toggled on or off individually. This conversion will happen over the course of the following PRs:

1. Fix CMO to pass the correct config directory to the monitoring-plugin so that the `*.patch.json` feature files are able to be loaded correctly (openshift/cluster-monitoring-operator#2981)
2. This PR which changes the monitoring-plugin backend to be feature driven and handle no feature flags so that the current monitoring-plugin features are enabled for backwards compatability
3. Change CMO to directly pass in the `alerting`, `targets`, `metrics` and `legacy-dashboards` feature flags to the monitoring-plugin. Ensure the `TestClusterMonitorConsolePlugin` test still passes
4. Adjust the monitoring-plugin to serve no features is no feature flags are passed in

## Summary by CodeRabbit

* **New Features**
  * Added console navigation, pages, and dev tabs for Alerting, Legacy Dashboards, Metrics, and Targets across standard and virtualization perspectives.
  * Enabled these observability features by default in local startup flows.

* **Bug Fixes**
  * Updated backend/console startup to use port **9443** and adjusted container startup config path.
  * Improved extension startup behavior by suppressing “no extensions” warnings.

* **Documentation**
  * Refreshed feature-flag, local development, and troubleshooting guidance, including updated startup commands and port instructions.


etmurasaki:
e2e-perses was run locally against a running cluster with COO already installed and it passed except scenarios involving bug https://redhat.atlassian.net/browse/OU-1466. Other than that, it was working fine.

pkg/plugin_handler.go
```diff
@@ -34,7 +34,10 @@ func manifestHandler(cfg *Config) http.HandlerFunc {
 
 func patchManifest(baseManifestData []byte, cfg *Config) []byte {
 	if len(cfg.Features) == 0 {
-		return baseManifestData
+		cfg.Features[Alerting] = true
```
jgbernalp: I know from config this wont be nil, but maybe from tests could be, should we initialize the features map?
```suggestion
        cfg.Features
  = make(map[Feature]bool)
		cfg.Features[Alerting] = true
```

pkg/plugin_handler.go
jgbernalp: Do we still need the dev mode which clears the extensions? If possible it would be better to remove it and let the dev environment script to initialize the features

config/alerting.patch.json
```diff
@@ -0,0 +1,323 @@
+[
+  {
+    "op": "add",
+    "path": "/extensions/1",
```
jgbernalp: can we change the patches to:

```suggestion
    "path": "/extensions/0",
```

So they can work even if there are no previous elements on the target json array?

jgbernalp:
I got an error while trying to start the backend locally:

```
ERRO[0000] cannot patch base manifest file               error="error in add for path: '/extensions/1': Unable to access invalid index: 1: invalid index referenced" module=manifest
```
this might be caused by trying to add an element to an empty array in the json.

config/alerting.patch.json
PeterYurkovich: Fixed in most recent commit

pkg/plugin_handler.go
PeterYurkovich: Fixed in most recent commit. The clear-extensions / dev was actually what was adding in the COO redux store, so I create a new file that just does that and is patched in when one of those features is enabled

pkg/plugin_handler.go
PeterYurkovich: Fixed, moved to a different file though since it makes sense to have all of the modifications to the cfg happen higher up in the server starting

pkg/plugin_handler.go
```diff
@@ -33,21 +33,20 @@ func manifestHandler(cfg *Config) http.HandlerFunc {
...
 
+	if cfg.Features[Alerting] || cfg.Features[LegacyDashboards] || cfg.Features[Metrics] || cfg.Features[Targets] {
+		patchedManifest = performPatch(patchedManifest, filepath.Join(cfg.ConfigPath, "monitoring-plugin.patch.json"))
+	}
+
+	if cfg.Features[Incidents] || cfg.Features[ClusterHealthAnalyzer] || cfg.Features[PersesDashboards] || cfg.Features[AcmAlerting] {
+		patchedManifest = performPatch(patchedManifest, filepath.Join(cfg.ConfigPath, "monitoring-console-plugin.patch.json"))
+	}
+
 	for feature := range cfg.Features {
```
jgbernalp: maybe this is a pre existing issue, but it seems the feature iteration to perform the patches is not deterministic. Should we assign some order to the features and sort before iterating?. Or is fine and we trust the console to resolve the extensions even if they are not in order. Not sure if there will be issues with some routes that might match before others.

PeterYurkovich: Routes are handled via react router, so determinism is influenced by that not the extension load order. It's not really an issue to add determinism though, is there a specific order we should do? Maybe:

- monitoring-plugin
- alerting
- metrics
- dashboards-legacy
- targets
- monitoring-console-plugin
- acm-alerting
- incidents
- dashboards-perses

Makefile
```diff
@@ -3,7 +3,8 @@ PLATFORMS   ?= linux/arm64,linux/amd64
 ORG         ?= openshift-observability-ui
 PLUGIN_NAME ?=monitoring-plugin
 IMAGE       ?= quay.io/${ORG}/${PLUGIN_NAME}:${VERSION}
-FEATURES    ?=incidents,perses-dashboards,dev-config
+MONITORING_FEATURES    ?=alerting,targets,legacy-dashboards,metrics
+ALL_FEATURES    ?=alerting,targets,legacy-dashboards,metrics,incidents,perses-dashboards
```
jgbernalp:
```suggestion
ALL_FEATURES    ?=$(MONITORING_FEATURES),incidents,perses-dashboards
```

PeterYurkovich: fixed in latest commit

pkg/plugin_handler.go
PeterYurkovich: fixed in latest commit

PeterYurkovich:
/override ci/prow/okd-scos-images

Image pull is failing in CI, other image job is succeeding, so going to override this one

PR #1035, opened 2026-07-06
NO-JIRA: fixing unit-test severity sort, variable template regex, and test corrections.

etmurasaki:
## Summary
Fix all failing unit tests (13 suites, 241 tests now passing).

### Changes
**`src/components/utils.ts`** — Fix `severitySort` comparison direction
The rank-based refactor accidentally used `rankB - rankA` (descending), but the function's contract is an ascending comparator (higher severity = positive return value). This is required by `directedSort`, which applies the user-chosen direction on top. Changed to `rankA - rankB`.

**`src/components/dashboards/legacy/variable-utils.ts`** — Fix `$__range` regex collision
The regex `\$__range` was matching inside `$__range_ms` and `$__range_s`, causing substitutions like `foo[1800s_ms]` instead of `foo[1800000]`. Added a negative lookahead `(?![a-zA-Z0-9_])` to prevent partial matches against longer variable names.

**`src/components/dashboards/legacy/variable-utils.spec.ts`** — Add SDK mock
The test transitively imports `@openshift-console/dynamic-plugin-sdk` (via `utils.ts`), which tries to resolve internal console aliases and PatternFly CSS. Added `jest.mock(...)` matching the pattern used by other test files.

**`src/components/Incidents/processAlerts.spec.ts`** — Fix incorrect test timestamp
The "should mark alert as firing if ended less than 10 minutes ago" test used a timestamp 14 minutes ago and assumed chart padding would shift the resolution check. However, resolution is intentionally determined from original (pre-padding) timestamps (as documented by the sibling test). Fixed the timestamp to 9 minutes ago (`nowSeconds - 540`).

**`jest.config.js` + `jest.styleMock.js`** — Mock CSS imports

Assited by Claude Code - Opus 4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alert state handling so recently ended alerts are classified more accurately.
  * Fixed dashboard variable matching to avoid accidental partial replacements inside longer strings.
  * Adjusted severity ordering checks to reflect the correct sort behavior.

### DIFF
--- a/web/src/components/dashboards/legacy/variable-utils.ts
+++ b/web/src/components/dashboards/legacy/variable-utils.ts
@@ -5,7 +5,9 @@ import { ALL_NAMESPACES_KEY } from '../../utils';
 const intervalVariableRegExps = ['__interval', '__rate_interval', '__auto_interval_[a-z]+'];
 
 export const isIntervalVariable = (itemKey: string): boolean =>
-  _.some(intervalVariableRegExps, (re) => itemKey?.match(new RegExp(`\\$${re}`, 'g')));
+  _.some(intervalVariableRegExps, (re) =>
+    itemKey?.match(new RegExp(`\\$${re}(?![a-zA-Z0-9_])`, 'g')),
+  );
 
 export type Variable = {
   isHidden?: boolean;
@@ -62,7 +64,7 @@ export const evaluateVariableTemplate = (
 
   let result = template;
   _.each(allVariables, (v, k) => {
-    const re = new RegExp(`\\$${k}`, 'g');
+    const re = new RegExp(`\\$${k}(?![a-zA-Z0-9_])`, 'g');
     if (result.match(re)) {
       if (v.isLoading) {
         result = undefined;
--- a/web/src/components/Incidents/processAlerts.spec.ts
+++ b/web/src/components/Incidents/processAlerts.spec.ts
@@ -275,9 +275,9 @@ describe('convertToAlerts', () => {
     });
 
     it('should mark alert as firing if ended less than 10 minutes ago', () => {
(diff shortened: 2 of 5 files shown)


web/src/components/utils.ts
```diff
@@ -152,7 +152,7 @@ export const severitySort = (
 
   const rankA = rank.get(severityA) ?? -1;
   const rankB = rank.get(severityB) ?? -1;
-  if (rankA !== rankB) return rankB - rankA;
+  if (rankA !== rankB) return rankA - rankB;
```
jgbernalp: I think this will affect the silences page, we need to adjust there. The alerts and alert rules add the sort direction so this is not an issue there, but silences will now show lowest severity first. It should be highest severity first.

web/src/components/dashboards/legacy/variable-utils.ts
```diff
@@ -62,7 +62,7 @@ export const evaluateVariableTemplate = (
 
   let result = template;
   _.each(allVariables, (v, k) => {
-    const re = new RegExp(`\\$${k}`, 'g');
+    const re = new RegExp(`\\$${k}(?![a-zA-Z0-9_])`, 'g');
```
jgbernalp, line 67: Line 8 needs the same adjustment

web/src/components/utils.ts
etmurasaki: I am reverting this change and instead, fix the testing expected results.

web/src/components/dashboards/legacy/variable-utils.ts
etmurasaki: done

PR #1036, opened 2026-07-07
NO-JIRA: fix unit test severitySort and incidents

etmurasaki:
image


PeterYurkovich:
/retitle NO-JIRA: fix unit test severitySort and incidents

PR #1038, opened 2026-07-07
OU-1389: restructure folders

PeterYurkovich:
This PR looks to update the folder structure of the proposed monitoring-plugin monorepo setup: https://github.com/observability-ui/harness/pull/4/changes

This PR performs as minimal changes as possible to the actual files, and only looks to move the files and create the new folder structure. Further work to do things like split files, refactor to fit the new structure, ect. are left to future PRs. 

This PR is independant of #1034 and they can be merged in either order

## Summary by CodeRabbit

* **New Features**
  * Added alert grouping and label filtering, including links between alerts and agentic runs.
  * Added Perses dashboard creation, editing, importing, migration, datasource support, and time-series visualization.
  * Added reusable table filtering, sorting, pagination, loading, error, and empty-state experiences.
  * Added shared incident models and improved incident charting and processing.
  * Added feature-flagged plugin manifest updates and expanded server validation.
* **Refactor**
  * Consolidated shared UI components, constants, hooks, and page integrations.
* **Tests**
  * Expanded coverage for server security, alerts, incidents, dashboards, formatting, and table utilities.


PeterYurkovich:
All coderabbit nitpick's are explicitly out of this PR's scope of work as stated in the description

web/src/shared/store/reducers.ts
```diff
@@ -2,7 +2,7 @@ import * as _ from 'lodash-es';
 
 import { ActionType, ObserveAction } from './actions';
 
-import { MONITORING_DASHBOARDS_VARIABLE_ALL_OPTION_KEY } from '../components/dashboards/legacy/utils';
+import { MONITORING_DASHBOARDS_VARIABLE_ALL_OPTION_KEY } from '../../features/legacy-dashboards/utils/utils';
```
jgbernalp, line 5: for a follow up PR. but we are importing here from features. Maybe it would be better to put this into constants.

web/src/shared/store/reducers.ts
```diff
@@ -11,7 +11,7 @@ import {
   QueryStructure,
 } from './store';
 import { produce } from 'immer';
-import { applySilences } from '../components/alerting/AlertUtils';
+import { applySilences } from '../../features/alerts/components/AlertUtils';
```
jgbernalp, line 14: same here it seems we are importing from features. Maybe we could move this into store in a follow up PR

web/src/shared/store/store.ts
```diff
@@ -1,15 +1,15 @@
...
-import { Silences } from '../components/types';
+import { Silences } from '../types/types';
 import {
   DaysFilters,
   IncidentSeverityFilters,
   IncidentStateFilters,
-} from '../components/Incidents/model';
-import { Variable } from '../components/dashboards/legacy/legacy-variable-dropdowns';
+} from '../../features/incidents/types/model';
+import { Variable } from '../../features/legacy-dashboards/components/legacy-variable-dropdowns';
```
jgbernalp, line 12: same case here.

web/src/shared/utils/utils.ts
```diff
@@ -15,9 +15,9 @@ import fuzzy from 'fuzzysearch';
 import * as _ from 'lodash-es';
 import { murmur3 } from 'murmurhash-js';
 
-import { QueryParams } from './query-params';
-import { AlertSource, MonitoringResource, Target, TimeRange } from './types';
-import { AggregatedAlert } from './alerting/AlertsAggregates';
+import { QueryParams } from '../constants/query-params';
+import { AlertSource, MonitoringResource, Target, TimeRange } from '../types/types';
+import { AggregatedAlert } from '../../features/alerts/pages/alerts-page/AlertsAggregates';
```
jgbernalp, line 20: maybe move this to types?

web/src/shared/components/query-browser/query-browser.tsx
```diff
@@ -53,46 +53,46 @@ import {
...
   parsePrometheusDuration,
-} from './console/console-shared/src/datetime/prometheus';
-import { getObserveState } from './hooks/usePerspective';
+} from '../../../shared/console/console-shared/src/datetime/prometheus';
+import { getObserveState } from '../../hooks/usePerspective';
 import './query-browser.scss';
-import { GraphUnits } from './metrics/units';
-import { DataTestIDs } from './data-test';
-import { useMonitoring } from '../hooks/useMonitoring';
+import { GraphUnits } from '../../../features/metrics/utils/units';
```
jgbernalp, line 93: same here

web/src/shared/components/dashboard-dropdown.tsx
```diff
@@ -13,8 +13,8 @@ import type { FC } from 'react';
 import { memo, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 
-import { SingleTypeaheadDropdown } from '../../console/utils/single-typeahead-dropdown';
-import { CombinedDashboardMetadata } from '../perses/hooks/useDashboardsData';
+import { SingleTypeaheadDropdown } from '../console/utils/single-typeahead-dropdown';
+import { CombinedDashboardMetadata } from '../../features/perses-dashboards/hooks/useDashboardsData';
```
jgbernalp, line 17: probably we are not using this shared dashboard dropdown anymore...

jgbernalp:
for a follow up. We probably need to update the `AGENTS.md` after these changes.

web/src/features/alerts/pages/alerts-page/AggregateAlertTableRow.tsx
jgbernalp, line 1: I thought the pages folder only contained pages, are we including also the components of those pages? should they be under feature/<feature>/components?

PeterYurkovich: I was thinking that the feature/components is the components that are shared across the feature, while feature/page/page-x/*.tsx would have component files that are directly related to that page. In this case the AggregateAlertTableRow is only used on the alerts-page, so while it could live under feature/components, putting it next to the only file that uses it makes more sense to me IMO.

PeterYurkovich:
Agree with all the locations that could have files split up and individual items moved to the shared folder. I just wanted to get the files moved in one PR and the address the rest in a follow up.

Also agree that the documentation needs to be updated, I will also do that in a follow up PR

web/src/features/alerts/pages/alerts-page/AggregateAlertTableRow.tsx
PeterYurkovich: I'm fine if we want to keep the feature/pages as only pages, it doesn't matter that much to me, I just thought that this structure helps with the mental overhead. Since all the pages files are named the same thing the only thing the user needs to check is the file name and location to understand that files relationship to all other files. If it is located in /feature/components its a shared component across the single feature, if it is /shared/components its a shared component across multiple features. If it is located in /feature/pages/page-x/ then it is a component related to a single page. And if the filename ends with `*Page.tsx` then you know it is the page component

PR #1020, opened 2026-06-30
NO-JIRA: feat: keep current data while loading new data in perses dashboards

jgbernalp:
Keeps the current charts data while loading new data to avoid flickering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dashboard and tool pages so previously loaded data stays visible while new results are loading.
  * Reduced flicker and empty states when switching filters, routes, or query parameters.

PR #1043, opened 2026-07-08
NO-JIRA: [release-coo-ocp-4.19] feat: keep current data while loading new data in perses dashboards

jgbernalp:
manual cherry pick of: https://github.com/openshift/monitoring-plugin/pull/1020


etmurasaki:
tested and could see clearly this change:
before:
image

after: there are spinning / loading icons in each panel
image

etmurasaki:
tested other scenarios as regression

PR #1047, opened 2026-07-08
CNV-80608: management: add update alert rule APIs

sradco:
Add PATCH /api/v1/alerting/rules for bulk
update of platform and user-defined alert
rules with drop/restore, label overrides,
and per-rule update support.

## Summary by CodeRabbit

* **New Features**
  * Added a bulk alert-rule update API (**PATCH** `/rules`) with **per-rule** results.
  * Supports label overrides (including removal via `null`/empty), classification overrides, and alert enable/disable with drop/restore.
* **Bug Fixes**
  * Improved management enforcement for GitOps- and operator-managed rules, including clearer per-rule outcomes for mixed success, not-found, and invalid IDs.
* **Tests**
  * Added unit and end-to-end coverage for bulk updates, drop/restore, classification behavior, and request validation limits/error cases.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -12,6 +12,52 @@ servers:
 
 paths:
   /rules:
+    patch:
+      operationId: BulkUpdateAlertRules
+      summary: Bulk update alert rules
+      description: >
+        Updates one or more alert rules by their stable IDs. Each rule is
+        updated independently; per-rule status is returned in the response
+        so partial success is visible to the caller.
+        Supports label overrides, drop/restore toggles (platform rules only),
+        and classification label updates.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/BulkUpdateAlertRulesRequest"
+      responses:
+        "200":
+          description: Update results (may include per-rule errors)
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/BulkUpdateAlertRulesResponse"
+        "400":
+          description: Invalid request body
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "401":
+          description: Missing or invalid authorization token
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "413":
+          description: Request body exceeds the 1 MB limit
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
+        "500":
+          description: Unexpected server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
     delete:
       operationId: BulkDeleteUserDefinedAlertRules
       summary: Bulk delete user-defined alert rules
@@ -218,6 +264,95 @@ components:
           items:
             $ref: "#/components/schemas/DeleteAlertRuleResult"
           description: Per-rule deletion results.
+
+    AlertRuleClassificationUpdate:
+      type: object
+      description: >
+        Partial update for alert rule classification labels.
+        Each field supports three states: omitted (leave unchanged),
+        null (clear the override), or a string value (set the override).
+        The three-state semantics require a custom JSON decoder; the Go
+        type AlertRuleClassificationPatch is used at runtime instead of
+        the generated struct.
+      x-go-type: AlertRuleClassificationPatch
+      properties:
+        openshift_io_alert_rule_component:
+          type: string
+          nullable: true
+          description: Component classification label override.
+        openshift_io_alert_rule_layer:
+          type: string
+          nullable: true
+          description: Layer classification label override.
+        openshift_io_alert_rule_component_from:
+          type: string
+          nullable: true
+          description: Dynamic component source label key.
+        openshift_io_alert_rule_layer_from:
+          type: string
+          nullable: true
+          description: Dynamic layer source label key.
+
+    BulkUpdateAlertRulesRequest:
+      type: object
+      required:
+        - ruleIds
+      properties:
+        ruleIds:
+          type: array
+          minItems: 1
+          maxItems: 100
+          items:
+            type: string
+          description: List of stable alert rule IDs to update (at most 100 per request).
+        labels:
+          type: object
+          additionalProperties:
+            type: string
+            nullable: true
+          description: >
+            Label key/value pairs to set. A null or empty-string value removes
+            the label. Omitting this field leaves existing labels unchanged.
+        alertingRuleEnabled:
+          type: boolean
+          nullable: true
+          description: >
+            When false, drops the alert rule via an AlertRelabelConfig Drop
+            action — the rule no longer appears in Prometheus query results.
+            When true, restores a previously dropped rule.
+            Only supported for platform alert rules.
+        classification:
+          $ref: "#/components/schemas/AlertRuleClassificationUpdate"
+
+    UpdateAlertRuleResult:
+      type: object
+      required:
+        - id
+        - statusCode
+      properties:
+        id:
+          type: string
+          description: The stable alert rule ID that was processed.
+        statusCode:
+          type: integer
+          format: int32
+          minimum: 100
+          maximum: 599
+          description: HTTP status code for this rule's update result.
+        message:
+          type: string
+          description: Error message if update failed; omitted on success.
+
+    BulkUpdateAlertRulesResponse:
+      type: object
+      required:
+        - rules
+      properties:
+        rules:
+          type: array
+          items:
+            $ref: "#/components/schemas/UpdateAlertRuleResult"
+          description: Per-rule update results.
 
     ErrorResponse:
       type: object
--- a/internal/managementrouter/alert_rule_bulk_update.go
+++ b/internal/managementrouter/alert_rule_bulk_update.go
@@ -1,0 +1,154 @@
+package managementrouter
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/openshift/monitoring-plugin/pkg/management"
+)
+
+func (hr *httpRouter) BulkUpdateAlertRules(w http.ResponseWriter, req *http.Request) {
+	req.Body = http.MaxBytesReader(w, req.Body, maxRequestBodyBytes)
+
+	body, err := io.ReadAll(req.Body)
+	if err != nil {
+		writeError(w, http.StatusRequestEntityTooLarge, "request body too large")
+		return
+	}
+
+	// BulkUpdateAlertRulesRequest.Classification is typed as
+	// *AlertRuleClassificationPatch (via x-go-type in the spec), so the
+	// three-state omitted/null/string semantics are preserved on decode.
+	var payload BulkUpdateAlertRulesRequest
+	if err := json.Unmarshal(body, &payload); err != nil {
+		writeError(w, http.StatusBadRequest, "invalid request body: "+err.Error())
+		return
+	}
+
+	if len(payload.RuleIds) == 0 {
+		writeError(w, http.StatusBadRequest, "ruleIds is required")
+		return
+	}
+	if len(payload.RuleIds) > maxBulkUpdateRuleIds {
+		writeError(w, http.StatusBadRequest, "ruleIds exceeds maximum of 100")
+		return
+	}
+
+	if payload.AlertingRuleEnabled == nil && payload.Labels == nil && payload.Classification == nil {
+		writeError(w, http.StatusBadRequest, "one of alertingRuleEnabled (toggle drop/restore) or labels (set/unset) or classification is required")
+		return
+	}
+	if payload.AlertingRuleEnabled != nil && (payload.Labels != nil || payload.Classification != nil) {
+		writeError(w, http.StatusBadRequest, "alertingRuleEnabled cannot be combined with labels or classification in the same request")
+		return
+	}
+
+	var haveToggle bool
+	var enabled bool
+	if payload.AlertingRuleEnabled != nil {
+		enabled = *payload.AlertingRuleEnabled
+		haveToggle = true
+	}
+
+	results := make([]UpdateAlertRuleResult, 0, len(payload.RuleIds))
+
+	for _, rawId := range payload.RuleIds {
+		id := strings.TrimSpace(rawId)
+		if id == "" {
+			msg := "ruleId is empty or whitespace-only"
+			results = append(results, UpdateAlertRuleResult{
+				Id:         rawId,
+				StatusCode: int32(http.StatusBadRequest),
+				Message:    &msg,
+			})
+			continue
+		}
+
+		if haveToggle {
+			var err error
+			if !enabled {
+				err = hr.managementClient.DropAlertRule(req.Context(), id)
+			} else {
+				err = hr.managementClient.RestoreAlertRule(req.Context(), id)
+			}
+			if err != nil {
+				status, message := parseError(err)
+				results = append(results, UpdateAlertRuleResult{
+					Id:         id,
+					StatusCode: int32(status),
+					Message:    &message,
+				})
+				continue
+			}
+			results = append(results, UpdateAlertRuleResult{
+				Id:         id,
+				StatusCode: int32(http.StatusNoContent),
+			})
+			continue
+		}
+
+		if payload.Classification != nil {
+			cl := payload.Classification
+			update := management.UpdateRuleClassificationRequest{RuleId: id}
+			if cl.ComponentSet {
+				update.Component = cl.Component
+				update.ComponentSet = true
+			}
+			if cl.LayerSet {
+				update.Layer = cl.Layer
+				update.LayerSet = true
+			}
+			if cl.ComponentFromSet {
+				update.ComponentFrom = cl.ComponentFrom
+				update.ComponentFromSet = true
+			}
+			if cl.LayerFromSet {
+				update.LayerFrom = cl.LayerFrom
+				update.LayerFromSet = true
+			}
+
+			if update.ComponentSet || update.LayerSet || update.ComponentFromSet || update.LayerFromSet {
+				if err := hr.managementClient.UpdateAlertRuleClassification(req.Context(), update); err != nil {
+					status, message := parseError(err)
+					results = append(results, UpdateAlertRuleResult{
+						Id:         id,
+						StatusCode: int32(status),
+						Message:    &message,
+					})
+					continue
+				}
+			}
+		}
+
+		if payload.Labels != nil {
+			newRuleId, err := hr.managementClient.UpdateAlertRuleLabels(req.Context(), id, *payload.Labels)
+			if err != nil {
+				status, message := parseError(err)
+				results = append(results, UpdateAlertRuleResult{
+					Id:         id,
+					StatusCode: int32(status),
+					Message:    &message,
+				})
+				continue
+			}
+			results = append(results, UpdateAlertRuleResult{
+				Id:         newRuleId,
+				StatusCode: int32(http.StatusNoContent),
+			})
+			continue
+		}
+
+		results = append(results, UpdateAlertRuleResult{
+			Id:         id,
+			StatusCode: int32(http.StatusNoContent),
+		})
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(BulkUpdateAlertRulesResponse{Rules: results}); err != nil {
+		log.WithError(err).Warn("failed to encode bulk update response")
+	}
+}
--- a/internal/managementrouter/router.go
+++ b/internal/managementrouter/router.go
@@ -22,7 +22,10 @@ var log = logrus.WithField("module", "managementrouter")
 // maxRequestBodyBytes limits incoming request bodies to 1 MB across all handlers.
 const maxRequestBodyBytes = 1 << 20 // 1 MB
 
-const maxBulkDeleteRuleIds = 100
+const (
+	maxBulkDeleteRuleIds = 100
+	maxBulkUpdateRuleIds = 100
+)
 
 type httpRouter struct {
 	managementClient management.Client
--- a/pkg/management/delete_user_defined_alert_rule_by_id.go
+++ b/pkg/management/delete_user_defined_alert_rule_by_id.go
@@ -2,6 +2,7 @@ package management
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	osmv1 "github.com/openshift/api/monitoring/v1"
@@ -64,7 +65,7 @@ func (c *client) deletePlatformAlertRuleById(ctx context.Context, relabeled moni
 		return fmt.Errorf("failed to get AlertingRule %s: %w", arName, err)
 	}
 	if !found || ar == nil {
-		return &NotFoundError{Resource: "AlertingRule", Id: arName}
+		return c.deleteUserAlertRuleById(ctx, namespace, name, alertRuleId)
 	}
 	// Common preconditions for platform delete
 	if err := validatePlatformDeletePreconditions(ar); err != nil {
@@ -85,9 +86,20 @@ func (c *client) deletePlatformAlertRuleById(ctx context.Context, relabeled moni
 			AdditionalInfo: fmt.Sprintf("alert %q not found in AlertingRule %s", originalRule.Alert, arName),
 		}
 	}
-	ar.Spec.Groups = newGroups
-	if err := c.k8sClient.AlertingRules().Update(ctx, *ar); err != nil {
-		return fmt.Errorf("failed to update AlertingRule %s: %w", ar.Name, err)
+
+	if len(newGroups) == 0 {
+		if err := c.k8sClient.AlertingRules().Delete(ctx, ar.Name); err != nil {
+			return fmt.Errorf("failed to delete AlertingRule %s: %w", ar.Name, err)
+		}
+	} else {
+		ar.Spec.Groups = newGroups
+		if err := c.k8sClient.AlertingRules().Update(ctx, *ar); err != nil {
+			return fmt.Errorf("failed to update AlertingRule %s: %w", ar.Name, err)
+		}
+	}
+
+	if err := c.deleteAssociatedARC(ctx, k8s.ClusterMonitoringNamespace, name, alertRuleId); err != nil {
+		return fmt.Errorf("failed to clean up ARC for platform rule %s: %w", alertRuleId, err)
 	}
 	return nil
 }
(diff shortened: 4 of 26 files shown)


internal/managementrouter/router.go
```diff
@@ -100,3 +104,15 @@ func parseError(err error) (int, string) {
...
+	decoded, err := url.PathUnescape(raw)
+	if err != nil {
+		return "", fmt.Errorf("invalid %s encoding", name)
+	}
+	value := strings.TrimSpace(decoded)
+	if value == "" {
+		return "", fmt.Errorf("missing %s", name)
+	}
+	return value, nil
+}
```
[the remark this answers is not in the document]
sradco: Fixed: replaced `parseParam` with `strings.TrimSpace` in the bulk update handler, consistent with the delete handler.

pkg/management/delete_user_defined_alert_rule_by_id.go
```diff
@@ -119,14 +133,39 @@ func (c *client) deleteUserAlertRuleById(ctx context.Context, namespace, name, a
...
+}
+
+// deleteAssociatedARC removes the AlertRelabelConfig associated with an alert rule, if it exists.
+// This is best-effort: if the ARC does not exist or is GitOps-managed, it is silently skipped.
+func (c *client) deleteAssociatedARC(ctx context.Context, namespace, prName, alertRuleId string) error {
+	arcName := k8s.GetAlertRelabelConfigName(prName, alertRuleId)
+	arc, found, err := c.k8sClient.AlertRelabelConfigs().Get(ctx, namespace, arcName)
+	if err != nil || !found {
+		return nil
+	}
```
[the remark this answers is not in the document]
sradco: Fixed: split the `if err != nil || !found` into separate checks -- genuine API errors are now surfaced, not-found is still silently skipped.

api/openapi.yaml
```diff
@@ -219,6 +268,97 @@ components:
...
+            type: string
+            nullable: true
+          description: >
+            Label key/value pairs to set. A null or empty-string value removes
+            the label. Omitting this field leaves existing labels unchanged.
+        alertingRuleEnabled:
+          type: boolean
+          nullable: true
+          description: >
+            When false, drops (silences) the platform alert rule.
```
PeterYurkovich: I don't think it is accurate to state that dropping an alert silences it, they are two separate actions

internal/managementrouter/alert_rule_bulk_update.go
```diff
@@ -0,0 +1,223 @@
...
+		writeError(w, http.StatusRequestEntityTooLarge, "request body too large")
+		return
+	}
+
+	// BulkUpdateAlertRulesRequest.Classification is typed as
+	// *AlertRuleClassificationPatch (via x-go-type in the spec), so the
+	// three-state omitted/null/string semantics are preserved on decode.
+	var payload BulkUpdateAlertRulesRequest
+	if err := json.Unmarshal(body, &payload); err != nil {
+		writeError(w, http.StatusBadRequest, "invalid request body")
```
PeterYurkovich: can we include the error in the response as well so that rather than trying to guess at what went wrong the caller can directly make an adjustment

internal/managementrouter/alert_rule_bulk_update.go
```diff
@@ -0,0 +1,223 @@
...
+		writeError(w, http.StatusBadRequest, "ruleIds is required")
+		return
+	}
+	if len(payload.RuleIds) > maxBulkUpdateRuleIds {
+		writeError(w, http.StatusBadRequest, "ruleIds exceeds maximum of 100")
+		return
+	}
+
+	if payload.AlertingRuleEnabled == nil && payload.Labels == nil && payload.Classification == nil {
+		writeError(w, http.StatusBadRequest, "alertingRuleEnabled (toggle drop/restore) or labels (set/unset) or classification is required")
```
PeterYurkovich:
```suggestion
		writeError(w, http.StatusBadRequest, "one of alertingRuleEnabled (toggle drop/restore) or labels (set/unset) or classification is required")
```

internal/managementrouter/alert_rule_bulk_update.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	results := make([]UpdateAlertRuleResult, 0, len(payload.RuleIds))
+
+	for _, rawId := range payload.RuleIds {
+		id := strings.TrimSpace(rawId)
+		if id == "" {
+			msg := "missing ruleId"
+			results = append(results, UpdateAlertRuleResult{
+				Id:         rawId,
```
PeterYurkovich, line 62: Won't this just be a whitespace only string?

internal/managementrouter/alert_rule_bulk_update.go
```diff
@@ -0,0 +1,223 @@
...
+				StatusCode: int32(http.StatusBadRequest),
+				Message:    &msg,
+			})
+			continue
+		}
+
+		notAllowedEnabled := false
+		if haveToggle {
+			var derr error
```
PeterYurkovich: Other checks use err, I don't see any reason to have this one have a different name

internal/managementrouter/alert_rule_bulk_update.go
```diff
@@ -0,0 +1,223 @@
...
+	var haveToggle bool
+	var enabled bool
+	if payload.AlertingRuleEnabled != nil {
+		enabled = *payload.AlertingRuleEnabled
+		haveToggle = true
+	}
+
+	results := make([]UpdateAlertRuleResult, 0, len(payload.RuleIds))
+
+	for _, rawId := range payload.RuleIds {
```
PeterYurkovich, line 57: If we are looping through up to 100 alert rules I think rather than repeatedly trying to perform an update only to get a permission failure it would be best to either cache the permission failures or perform the permission checks ahead of time

pkg/management/update_platform_alert_rule.go
```diff
@@ -0,0 +1,417 @@
...
+
+	namespace := rule.Labels[k8s.PrometheusRuleLabelNamespace]
+	name := rule.Labels[k8s.PrometheusRuleLabelName]
+	nn := types.NamespacedName{Namespace: namespace, Name: name}
+
+	arcNamespace, err := c.arcNamespaceForRule(nn)
+	if err != nil {
+		return err
+	}
+	isPlatform := c.isPlatformManagedPrometheusRule(nn)
```
PeterYurkovich: In what circumstances do we want the `UpdatePlatformAlertRule` to be called with non-platform PrometheusRules?

pkg/management/update_platform_alert_rule.go
```diff
@@ -0,0 +1,417 @@
...
+	if err := validateGitOpsPreconditions(rule, prMeta); err != nil {
+		return err
+	}
+
+	originalRule, err := getOriginalPlatformRuleFromPR(prMeta, namespace, name, alertRuleId)
+	if err != nil {
+		return err
+	}
+
+	if v, ok := alertRule.Labels[managementlabels.AlertNameLabel]; ok {
```
PeterYurkovich, line 61: Rather than checking for specific immutable labels could we create a `validateProtectedLabels` function which takes in both alert rules and ensures that all protected labels are not mutated? Similar to `filterAndValidatePlatformLabelChanges` but not for filtering them out

pkg/management/update_platform_alert_rule.go
```diff
@@ -0,0 +1,417 @@
...
+		}
+	}
+
+	// AlertingRule CR path is only available for platform rules
+	if isPlatform {
+		arName := rule.Labels[managementlabels.AlertingRuleLabelName]
+		if arName == "" {
+			arName = defaultAlertingRuleName
+		}
+		ar, arFound, arErr := c.getAlertingRule(ctx, arName)
```
PeterYurkovich: Whats the difference between `ar` and `originalRule`

pkg/management/update_platform_alert_rule.go
```diff
@@ -0,0 +1,417 @@
...
+			return arErr
+		}
+		if arFound && ar != nil {
+			if gitOpsManaged, operatorManaged := k8s.IsExternallyManagedObject(ar); gitOpsManaged {
+				return &NotAllowedError{Message: "This alert is managed by GitOps; edit it in Git."}
+			} else if operatorManaged {
+				return c.applyLabelChangesViaAlertRelabelConfig(ctx, arcNamespace, alertRuleId, *originalRule, alertRule.Labels)
+			}
+			return c.updateAlertingRuleLabels(ctx, ar, originalRule.Alert, alertRuleId, alertRule.Labels, arName)
+		}
```
PeterYurkovich, line 80: Should `!arFound` be an error case?

pkg/management/label_utils.go
```diff
@@ -1,6 +1,20 @@
...
+var protectedLabels = map[string]bool{
+	managementlabels.AlertNameLabel: true,
+	k8s.AlertRuleLabelId:            true,
+}
+
+func isProtectedLabel(label string) bool {
+	return protectedLabels[label]
+}
 
 var validSeverities = map[string]bool{
```
PeterYurkovich, line 19: This may have already been gone over, but alertmanager doesn't have a set of valid severities, and user's can put whatever strings they want as a severity. Was it a design decision to only allow these 4?

internal/managementrouter/alert_rule_bulk_update.go
```diff
@@ -0,0 +1,223 @@
...
+					})
+					continue
+				}
+
+				var na *management.NotAllowedError
+				if errors.As(err, &na) {
+					updatedUserRule := currentRule
+					updatedUserRule.Labels = userLabels
+
+					newRuleId, err := hr.managementClient.UpdateUserDefinedAlertRule(req.Context(), id, updatedUserRule)
```
PeterYurkovich: Shouldn't we determine if an alert rule is a platform or user defined ar before just trying platform and then falling back to user on a not allowed error?

internal/managementrouter/alert_rule_bulk_update.go
```diff
@@ -0,0 +1,223 @@
...
+			continue
+		}
+
+		notAllowedEnabled := false
+		if haveToggle {
+			var derr error
+			if !enabled {
+				derr = hr.managementClient.DropPlatformAlertRule(req.Context(), id)
+			} else {
+				derr = hr.managementClient.RestorePlatformAlertRule(req.Context(), id)
```
PeterYurkovich: Is it not possible to drop/restore user ar's? For example if an ar is gitops or operator managed?

internal/managementrouter/alert_rule_bulk_update.go
```diff
@@ -0,0 +1,223 @@
...
+				update.ComponentFrom = cl.ComponentFrom
+				update.ComponentFromSet = true
+			}
+			if cl.LayerFromSet {
+				update.LayerFrom = cl.LayerFrom
+				update.LayerFromSet = true
+			}
+
+			if update.ComponentSet || update.LayerSet || update.ComponentFromSet || update.LayerFromSet {
+				if err := hr.managementClient.UpdateAlertRuleClassification(req.Context(), update); err != nil {
```
PeterYurkovich, line 113: I think looking more at the code it would be best to have the label update code and the drop/restore code also flow through a single function similar to `UpdateAlertRuleClassification` for the classification rather than being split into platform and user functions which are called in this http method. The difference between `user` and `platform` seems that it is being mixed in with whether something is gitops or operator managed. I think having a single point of entry for those actions which can perform all of the validation and route the actions to user or platform would help clarify the codepaths quite a bit

internal/managementrouter/alert_rule_bulk_update.go
sradco: Good point. 
I addressed in this update. 
I've introduced a unified `UpdateAlertRuleLabels` method in the management package that uses `isPlatformManagedPrometheusRule` to route internally (same pattern as `DeleteAlertRuleById` and `UpdateAlertRuleClassification`). The handler now simply calls:

```go
newRuleId, err := hr.managementClient.UpdateAlertRuleLabels(req.Context(), id, *payload.Labels)
```

No more error-type sniffing or fallback logic in the HTTP layer.

internal/managementrouter/alert_rule_bulk_update.go
sradco: I agree. Done in this push. 
I've introduced `UpdateAlertRuleLabels` as a unified entry point in the management package (similar to how `UpdateAlertRuleClassification` works). It:

1. Looks up the rule once
2. Uses `isPlatformManagedPrometheusRule` to determine the path
3. Routes to platform (ARC-based overlay) or user-defined (direct PrometheusRule mutation) internally

The HTTP handler now has no platform/user branching at all for labels — it's a single method call. The "try platform, fall back on NotAllowedError" pattern is completely gone.

pkg/management/label_utils.go
sradco: Yes, this is intentional. These are the four severities that the OpenShift platform monitoring stack uses and the console UI maps to specific visual treatments (colors, icons, routing behavior):

- `critical` — pages on-call
- `warning` — needs attention but not immediate
- `info` — informational / no action needed
- `none` — infrastructure signals (e.g. the Watchdog alert uses this severity)

Allowing arbitrary strings would produce alerts that the console can't categorize or display properly. Since this API is specifically for the monitoring-plugin/console UI, the restriction is deliberate.

pkg/management/update_platform_alert_rule.go
sradco: No. the fallback to the ARC path is intentional. 
Platform alerts always have a PrometheusRule (that's how Prometheus evaluates them), but they may not have an AlertingRule CR. 
The AlertingRule CR is an optional CMO resource that some (not all) platform alerting rules use.

When the AlertingRule CR doesn't exist (either not yet created, or the rule uses a different management mechanism), label overrides are applied via AlertRelabelConfig -  which is the universal fallback path that works for any platform rule regardless of whether an AlertingRule CR exists.

pkg/management/update_platform_alert_rule.go
sradco: The naming is admittedly misleading now. `UpdatePlatformAlertRule` handles label overrides for rules in platform-managed PrometheusRules via ARC, but with the new `UpdateAlertRuleLabels` unified entry point, callers no longer need to know this distinction — the routing happens internally.

The method still exists as the internal platform path, but it's no longer called directly from the HTTP handler. The unified method calls it after determining the rule is platform-managed.

pkg/management/update_platform_alert_rule.go
sradco: - \`ar\` is the AlertingRule CR (an \`osmv1.AlertingRule\`) — the CMO-managed resource that defines how alerts are projected/managed. It holds the rule's groups/rules as the operator intends them.
- \`originalRule\` is the rule as it appears in the source PrometheusRule resource — this is the ground-truth definition (alert name, expr, labels) before any relabeling is applied.

We need both because the AlertingRule CR is where we write label changes (if it exists and is editable), but we derive the immutable alert name and original labels from the PrometheusRule (which is what Prometheus actually evaluates).

pkg/management/update_platform_alert_rule.go
sradco: Good idea. Currently we only have two protected labels (\`openshift_io_alert_name\` and \`openshift_io_alert_rule_id\`), and the check is just for the alert name immutability. A dedicated \`validateProtectedLabels\` function would be cleaner if the set grows. I'll add this as a follow-up improvement — for now the set is small enough that the explicit check works, but I agree it should be generalized.

internal/managementrouter/alert_rule_bulk_update.go
sradco: Good point about performance. In practice though, the permission check is embedded in the Drop/Restore operation itself (it's part of the \`arcNamespaceForRule\` call which is a simple namespace check — no API call). The actual K8s API calls only happen if the rule passes preconditions.

That said, with the new \`UpdateAlertRuleLabels\` unified method, we could add a pre-check batch optimization in a follow-up if profiling shows this is a bottleneck with large batches. For now, each iteration does one \`RelabeledRules().Get()\` (in-memory cache hit) + namespace check, which is O(1) per rule.

internal/managementrouter/alert_rule_bulk_update.go
sradco: Fixed — renamed to \`err\` in the updated code, consistent with the rest of the handler.

internal/managementrouter/alert_rule_bulk_update.go
sradco: Good catch — after \`strings.TrimSpace(rawId)\`, the result is either empty or a non-whitespace string, so the message was inaccurate. Updated to \`\"ruleId is empty or whitespace-only\"\` to be precise about what triggered it — the trimmed value was empty, meaning the original was either empty or whitespace-only.

internal/managementrouter/alert_rule_bulk_update.go
sradco: Already done — each per-rule result includes a \`message\` field with the error string when the status is non-2xx. The \`parseError(err)\` call extracts the error message and includes it in the response. For example:

\`\`\`json
{"rules": [{"id": "...", "statusCode": 405, "message": "This alert is managed by GitOps; edit it in Git."}]}
\`\`\`

The top-level 400 errors (invalid body, missing ruleIds, etc.) also include error messages in the response body.

api/openapi.yaml
sradco: You're right — updated the description. Drop prevents the alert from appearing in Prometheus results entirely (via a Drop relabel action), while silencing is an Alertmanager concept that suppresses notifications but keeps the alert visible. The API description now uses "toggle drop/restore" to be precise.

internal/managementrouter/alert_rule_bulk_update.go
sradco: Applied — the message now reads:
\`"one of alertingRuleEnabled (toggle drop/restore) or labels (set/unset) or classification is required"\`

internal/managementrouter/alert_rule_bulk_update.go
sradco: I've extended drop/restore to work for user-defined rules as well (when `ENABLE_USER_WORKLOAD_ARCS` is enabled). The infrastructure was already there - `arcNamespaceForRule` routes user rules to `openshift-user-workload-monitoring`, and the ARC creation/deletion logic is namespace-agnostic.

This covers the use cases you mentioned:
- Operator-managed user rules that can't be deleted
- GitOps-managed user rules where the PrometheusRule is reconciled externally

The functions are now renamed to `DropAlertRule`/`RestoreAlertRule` (no longer platform-specific), and the handler no longer has the `NotAllowedError` fallback pattern - it just reports errors directly.

pkg/management/update_platform_alert_rule.go
PeterYurkovich: After your update with the `UpdateAlertRuleLabels` the `isPlatform` is now a dead codepath. This function is always called when something is platform managed

PeterYurkovich (commented):
If there is an operator owned alert in a non-platform namespace is there any way to manage labels?

test/e2e/update_alert_rule_test.go
```diff
@@ -0,0 +1,245 @@
...
+	osmv1 "github.com/openshift/api/monitoring/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/monitoring-plugin/internal/managementrouter"
+	"github.com/openshift/monitoring-plugin/pkg/k8s"
+	"github.com/openshift/monitoring-plugin/pkg/managementlabels"
+	"github.com/openshift/monitoring-plugin/test/e2e/framework"
+)
+
+func TestUpdateAlertRule_DropRestore(t *testing.T) {
```
simonpasquier, line 25: Do we have tests verifying that requests with unauthorized users are denied?

simonpasquier: Ideally we need to cover
* User A has no permission to update PrometheusRule in any namespace
  * Request fails for namespace Y
* User B has permission to update PrometheusRule in namespace Y
  * Request succeeds for namespace Y
  * Request fails for namespace Z
* User C has permission to update PrometheusRule in any namespace
  * Request succeeds for namespace Y
  * Request succeeds for namespace Z

sradco:
> If there is an operator owned alert in a non-platform namespace is there any way to manage labels?

Unfortunately no. Users can only silence these alerts. They can then copy them and update to their needs.
This is a know limitation. 
I did add a knob, in case that CMO will decide to support ARC also for the non-platform stack.

test/e2e/update_alert_rule_test.go
simonpasquier: And it would be way way easier to review if these new scenarios are implemented in a separate PR (prior this one) for the existing API endpoints (create/delete),

test/e2e/helpers_test.go
```diff
@@ -9,12 +9,45 @@ import (
...
+	alertrule "github.com/openshift/monitoring-plugin/pkg/alert_rule"
+	"github.com/openshift/monitoring-plugin/pkg/k8s"
 	"github.com/openshift/monitoring-plugin/test/e2e/framework"
 )
 
 func strPtr(s string) *string { return &s }
 
+// findPlatformAlertRuleId discovers an existing platform alert rule from
+// openshift-monitoring and returns its computed alert rule ID.
+func findPlatformAlertRuleId(t *testing.T, f *framework.Framework, ctx context.Context) string {
```
simonpasquier: ctx should be the first parameter

test/e2e/update_alert_rule_test.go
sradco: I created https://github.com/openshift/monitoring-plugin/pull/1066 to add the above tests.

PeterYurkovich:
/override-sticky ci/prow/test-frontend-unit
I'll rebase the base branch to main to get new test later today

PeterYurkovich:
/override-sticky ci/prow/e2e-monitoring
I'll rebase the base branch to main to get new test later today

api/openapi.yaml
```diff
@@ -219,6 +268,95 @@ components:
             $ref: "#/components/schemas/DeleteAlertRuleResult"
           description: Per-rule deletion results.
 
+    AlertRuleClassificationUpdate:
```
simonpasquier, line 268: unrelated to the topic of this PR?

simonpasquier (commented):
It looks like the 2nd commit includes the addition of the update API endpoint but also changes related to alert classification. Is this expected? It makes the diff harder to review than required.

test/e2e/helpers_test.go
sradco: Done — reordered to `findPlatformAlertRuleId(ctx, t, f)` and `mustCreateRule(ctx, t, f, ...)`.

api/openapi.yaml
sradco: `AlertRuleClassificationUpdate` is intentionally part of this PR: the new `BulkUpdateAlertRules` endpoint accepts classification patches alongside labels and drop/restore, so the OpenAPI schema has to define that payload here.

Separately, I reverted a drive-by change on `DeleteAlertRuleResult.statusCode` (`format: int32` / min/max) that was unrelated to update APIs.

test/e2e/update_alert_rule_test.go
sradco: Covered in https://github.com/openshift/monitoring-plugin/pull/1066 (create/delete RBAC e2e). This PR is now rebased on top of that branch tip; the update-path RBAC cases in `update_alert_rule_test.go` follow the same pattern.

pkg/management/update_platform_alert_rule.go
sradco: Agreed for the Drop/Restore paths after `arcNamespaceForRule` succeeds — removed the redundant `isPlatformManagedPrometheusRule` guards there.

`UpdatePlatformAlertRule` still keeps `arcNamespaceForRule` as a guard because it remains on the exported `Client` interface and has a unit test for the non-platform rejection path (`TestUpdatePlatformAlertRule_UserRuleReturnsError`). Production bulk-update traffic goes through `UpdateAlertRuleLabels`, which already routes platform vs user before calling it.

sradco:
Rebased onto latest `main-alerts-management-api` + tip of #1066 (`e2e-rbac-tests`).

Addressed remaining review items:
- `ctx` first on e2e helpers
- removed redundant `isPlatform` checks in Drop/Restore after `arcNamespaceForRule`
- kept `AlertRuleClassificationUpdate` (needed by `BulkUpdateAlertRules`); reverted unrelated `DeleteAlertRuleResult.statusCode` int32 drive-by

Note: PR base stays `main-alerts-management-api` (fork branch `e2e-rbac-tests` is not available as an upstream base), so GitHub still shows the #1066 commit until that PR merges.

/retest

pkg/management/update_alert_rule_labels.go
```diff
@@ -0,0 +1,85 @@
...
+		return c.updatePlatformRuleLabels(ctx, alertRuleId, labels)
+	}
+
+	return c.updateUserRuleLabels(ctx, alertRuleId, rule, labels)
+}
+
+// updatePlatformRuleLabels applies label changes to a platform rule via ARC.
+// The platform path uses "" to signal "drop this label".
+func (c *client) updatePlatformRuleLabels(ctx context.Context, alertRuleId string, labels map[string]*string) (string, error) {
+	platformLabels := make(map[string]string, len(labels))
```
jgbernalp, line 35: Should this check for protected labels using `filterAndValidatePlatformLabelChanges`?

pkg/management/update_platform_alert_rule.go
```diff
@@ -0,0 +1,468 @@
...
+
+func filterAndValidatePlatformLabelChanges(labels map[string]string) (map[string]string, error) {
+	filtered := make(map[string]string)
+	for k, v := range labels {
+		if !isProtectedLabel(k) {
+			filtered[k] = v
+		}
+	}
+	for k, v := range filtered {
+		if k == managementlabels.AlertNameLabel {
```
jgbernalp: this can never match as is excluded in the isProtectedLabel check

pkg/management/update_alert_rule_labels.go
```diff
@@ -0,0 +1,85 @@
...
+		return "", &NotFoundError{Resource: "PrometheusRule", Id: alertRuleId}
+	}
+
+	sourceRule, err := getOriginalPlatformRuleFromPR(pr, namespace, name, alertRuleId)
+	if err != nil {
+		return "", err
+	}
+
+	userLabels := make(map[string]string, len(sourceRule.Labels))
+	for k, v := range sourceRule.Labels {
```
jgbernalp: `copyStringMap` can be used here

pkg/management/update_alert_rule_labels.go
sradco: Updated

pkg/management/update_platform_alert_rule.go
sradco: Removed

pkg/management/update_alert_rule_labels.go
sradco: Updated

pkg/management/update_alert_rule_labels.go
```diff
@@ -0,0 +1,87 @@
...
+	if !prFound {
+		return "", &NotFoundError{Resource: "PrometheusRule", Id: alertRuleId}
+	}
+
+	sourceRule, err := getOriginalPlatformRuleFromPR(pr, namespace, name, alertRuleId)
+	if err != nil {
+		return "", err
+	}
+
+	userLabels := copyStringMap(sourceRule.Labels)
```
jgbernalp, line 75: This merges the labels unfiltered as is done in other places with the `isProtectedLabel`

jgbernalp: @sradco this can be fixed in a follow up PR.

PR #1051, opened 2026-07-09
Fix for OCPBUGS-98164: CVE-2026-13676

germanparente:
## Summary
- Fixes CVE-2026-13676: security policy bypass via improper Unicode hostname canonicalization in `fast-uri` (CVSS 7.5)
- Adds `fast-uri` override pinned to `3.1.3` in `web/package.json` (preventive — current dependency tree uses `uri-js` via `ajv` 6.12.6, but the override ensures protection if a future dependency update introduces `fast-uri`)
- Follows established CVE fix pattern (npm overrides)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -124,7 +124,8 @@
     "http-errors": "2.0.1",
     "sass": {
       "immutable": "^4.3.7"
-    }
+    },
+    "fast-uri": "3.1.3"
   },
   "consolePlugin": {
     "name": "monitoring-plugin",


jgbernalp:
@germanparente was `npm install` executed here, i see no changes on the package-lock.json which means the override might not be needed

germanparente:
@jgbernalp you are right. It was preventive but the best would be to close this PR and mark jira as not a bug. Apologies for the noise and thanks a lot for your review.

germanparente:
@jgbernalp in fact, it's better to re-open. Since the CVE in 4.17 needs to be fixed, I would say this preventive commit could make sense. Do you agree ?

germanparente:
@jgbernalp could we discuss about this one ?

jgbernalp:
@germanparente I don't think this is needed even for prevention, the dependency is not present, not even as a transitive dependency. We can close the CVE marking that the vulnerable code is not present. Feel free to close this PR

germanparente:
ok. I will do. Thanks @jgbernalp  What is strange is that only 4.18 is exempted of this and 4.17 needs the fix.
So, the issue is that the 4.17 PR should skip validation on 4.18 one (closed as not a bug).

Closing

PR #1053, opened 2026-07-13
Fix for OCPBUGS-89698: CVE-2026-12143

rhdmalone:
Fix for CVE-2026-12143: form-data does not escape CR/LF/quote in multipart field names and filenames (CRLF injection)
updates form-data to 4.0.6 or higher and makes other associated changes

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -14043,16 +14043,16 @@
       }
     },
     "node_modules/form-data": {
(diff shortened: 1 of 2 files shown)


germanparente:
/label jira/skip-dependent-bug-check

since 5.0 is not using form-data except for dependencies and it's already 4.0.6

PR #1056, opened 2026-07-14
OU-1389: move components and utils into shared locations

PeterYurkovich:
This PR looks to move imports from other pages and features into shared locations. It also creates a new lint rule which enforces this behavior

## Summary by CodeRabbit

* **New Features**
  * Added shared severity badges and graph show/hide controls across monitoring views.
  * Added dashboard permission checks to control editing actions.
  * Added reusable table sorting, filtering, and URL-synchronized pagination support.

* **Improvements**
  * Standardized shared dashboard metadata, graph units, and table utilities.
  * Improved import boundaries and project validation to promote more consistent feature organization.

* **Testing**
  * Added coverage for import-boundary validation across feature and page structures.


web/src/features/legacy-dashboards/components/dashboard-skeleton-legacy.tsx
```diff
@@ -4,7 +4,7 @@ import { memo, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { Divider, PageSection, Split, SplitItem, Stack, StackItem } from '@patternfly/react-core';
-import { CombinedDashboardMetadata } from '../../perses-dashboards/hooks/useDashboardsData';
+import type { CombinedDashboardMetadata } from '../../../shared/types/types';
```
jgbernalp, line 7: In a follow up PR. We probably want to break away this shared type between the legacy dashboards and Perses.

web/src/features/perses-dashboards/pages/dashboard-page/dashboard-frame.tsx
```diff
@@ -1,7 +1,7 @@
 import { FC, ReactNode } from 'react';
 import { DashboardEmptyState } from '../../components/emptystates/DashboardEmptyState';
 import { DashboardHeader } from '../../components/dashboard-header';
-import { CombinedDashboardMetadata } from '../../hooks/useDashboardsData';
+import type { CombinedDashboardMetadata } from '../../../../shared/types/types';
```
jgbernalp, line 4: to avoid the long import navigation chain we can think about start using aliases. `@/shared/types/types`

PeterYurkovich: I was also thinking about this, and its actually one of the primary reasons I went with the ESLint plugin over writing a custom rule. There are other rules we can look to apply within the plugin, like `no-relative-packages` to enforce things like that. It would also be good to set an enforced import order so that we don't have imports moving up and down the list every time a change is made

PeterYurkovich:
/label qe-approved
e2e test validated

web/src/shared/components/table/hooks/useTableFilters.ts
jgbernalp, line 1: should this be in some `hooks` folder?

PeterYurkovich: Sure, I would still like to keep the table hooks bundled together though. Would you prefer `/shared/components/table/hooks/*` or `/shared/hooks/table/*`

jgbernalp: `/shared/components/table/hooks/*` if the table component is the only consumer

PeterYurkovich: Fixed in latest commit

PR #1061, opened 2026-07-15
OU-1389: don't match on feature_* webpack bundles

PeterYurkovich:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved endpoint routing so health, manifest, feature, and configuration requests only match their exact paths.
  * Prevented unintended matches for URLs that merely begin with these endpoint paths.


PeterYurkovich:
/label qe-approved
Change only relevant for dev server

PR #1073, opened 2026-07-21
OU-1389: update devspace setup

PeterYurkovich:
This PR looks to update devspace support for the new development and feature setup. It adds a new Makefile command (`make start-mcp-frontend`) which injects the correct package name into the webpack build and splits the devspace command to take over either the CMO or COO deployments. If the COO deployment is selected then the command will attempt to install COO if it doesn't already exist and create the `monitoring-console-plugin` deployment.

## Summary by CodeRabbit

* **New Features**
  * Enhanced Devspace development to support both Monitoring (MP/CMO) and COO modes with a selectable plugin workflow.
  * Added an MCP-specific Devspace feature bundle plus a dedicated `start-mcp-frontend` workflow.
  * Introduced automated COO/CMO install and scale helper scripts.

* **Documentation**
  * Reworked “Running using Devspace” with clearer MP vs COO/CMO vs COO flow and commands.

* **Bug Fixes**
  * Improved dynamic console plugin i18n namespace handling for local runs and Cypress.

* **Chores**
  * Updated the dev web-builder base image (Go version) and streamlined plugin-name handling.
  * Removed the legacy DevSpace deployment configuration and deprecated plugin-name update steps.


devspace.yaml
```diff
@@ -2,47 +2,68 @@ version: v2beta1
...
+        echo "==> Applying UIPlugin CR with all features enabled..."
+        ./hack/create-uiplugin.sh
+        echo "==> Scaling down COO..."
+        ./hack/scale-down-coo.sh
+        start_dev app-mcp
+      else
+        echo "==> Scaling down CMO..."
+        ./hack/scale-down-cmo.sh
+        start_dev app-mp
+      fi
```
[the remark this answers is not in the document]
PeterYurkovich: Devspace only allows the selection of those 2 values. This is not an issue

hack/install-coo.sh
```diff
@@ -0,0 +1,51 @@
...
+until kubectl get csv -n openshift-cluster-observability-operator \
+    -o jsonpath='{.items[0].status.phase}' 2>/dev/null | grep -q "^Succeeded$"; do
+    if [[ ${SECONDS_ELAPSED} -ge 300 ]]; then
+        echo "ERROR: Timed out waiting for COO CSV to succeed."
+        exit 1
+    fi
+    sleep 5
+    SECONDS_ELAPSED=$((SECONDS_ELAPSED + 5))
+    echo "  Waiting... (${SECONDS_ELAPSED}s / 300s)"
+done
```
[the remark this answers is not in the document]
PeterYurkovich: The `openshift-cluster-observability-operator` namespace is created exclusively for this operator, so it is a reasonable assumption to make

PeterYurkovich:
/label qe-approved
Only development environment change

hack/scale-down-cmo.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+echo "Patching ClusterVersion to disable CMO management of monitoring-plugin..."
+oc patch clusterversion version --type json \
+  -p "$(cat ./web/cypress/fixtures/cmo/disable-monitoring.yaml)"
+
+echo "Scaling down cluster-monitoring-operator in openshift-monitoring..."
+kubectl scale --replicas=0 -n openshift-monitoring deployment/cluster-monitoring-operator
```
jgbernalp, line 9: should we wait until this action is completed, maybe the dev space will assume this is already down and replace the image that still being reconciled by the operator

Makefile
```diff
@@ -31,6 +32,10 @@ build-frontend:
 start-frontend:
 	cd web && npm run start
 
+.PHONY: start-mcp-frontend
+start-mcp-frontend:
+	cd web && CONSOLE_PLUGIN_NAME=monitoring-console-plugin npm run start
```
jgbernalp, line 37: should we set the I18N_NAMESPACE here as well?

web/webpack.config.ts
```diff
@@ -100,7 +104,9 @@ const config: Configuration = {
     }),
     new DefinePlugin({
       // Build-time injection of proxy path for config module
-      PERSES_PROXY_BASE_URL: JSON.stringify('/api/proxy/plugin/monitoring-console-plugin/perses'),
+      PERSES_PROXY_BASE_URL: JSON.stringify(
+        `/api/proxy/plugin/${process.env.CONSOLE_PLUGIN_NAME ?? pkg.consolePlugin.name}/perses`,
```
jgbernalp: the perses proxy does not exist in "MCO mode"

hack/install-coo.sh
```diff
@@ -0,0 +1,51 @@
...
+  channel: fast
+  name: cluster-observability-operator
+  source: redhat-operators
+  sourceNamespace: openshift-marketplace
+EOF
+
+echo "Waiting for COO CSV to reach Succeeded phase (timeout 5m)..."
+SECONDS_ELAPSED=0
+until kubectl get csv -n openshift-cluster-observability-operator \
+    -o jsonpath='{.items[0].status.phase}' 2>/dev/null | grep -q "^Succeeded$"; do
```
jgbernalp: should we search for the actual status that marks the installation as successful rather than assuming is the first one? or is it always the first one.

devspace.yaml
```diff
@@ -2,47 +2,68 @@ version: v2beta1
...
+
   purge:
     run: |-
       stop_dev --all
-      scale_up_cmo
-
-deployments: $( [ ${USE_HELM} == "y" ] && cat devspace-deployment.yaml || echo "app:")
```
jgbernalp, line 29: Is the `devspace-deployment.yaml` needed or can it be removed as it won't be used by removing this line.

hack/scale-down-cmo.sh
PeterYurkovich: Done

Makefile
PeterYurkovich: The initial answer to this was no due to the naming of our translation files, but you helped me realize that we can actually improve our build process more. The initial change in the webpack config that this PR contained already performed the first of 2 actions that the `update-plugin-name.sh` script performed, and I added the second as well. Now rather than need to run that script before building the `monitoring-console-plugin` in our dockerfiles we can just set the ENV variable and let webpack take over handling it

I've also combined the two ENV variable to just the CONSOLE_PLUGIN_NAME and have the old usage of I18N_NAMESPACE in the webpack file be constructed from the CONSOLE_PLUGIN_NAME.

When we create the next konflux update including these changes we will need to remove the script running and just pass in the ENV

web/webpack.config.ts
PeterYurkovich: Thats fair, I've reverted this change

hack/install-coo.sh
PeterYurkovich: Done. It would be a very odd situation to not have COO installed in `openshift-cluster-observability-operator` but have another operator installed in that namespace, but no harm in looking for it directly

devspace.yaml
PeterYurkovich: You are correct, I've remove the file.

scripts/start-console.sh
```diff
@@ -19,7 +19,7 @@ BRIDGE_K8S_MODE_OFF_CLUSTER_THANOS=$(oc -n openshift-config-managed get configma
 BRIDGE_K8S_MODE_OFF_CLUSTER_ALERTMANAGER=$(oc -n openshift-config-managed get configmap monitoring-shared-config -o jsonpath='{.data.alertmanagerPublicURL}')
 BRIDGE_K8S_AUTH_BEARER_TOKEN=$(oc whoami --show-token 2>/dev/null)
 BRIDGE_USER_SETTINGS_LOCATION="localstorage"
-BRIDGE_I18N_NAMESPACES=plugin__monitoring-plugin
+BRIDGE_I18N_NAMESPACES="plugin__${npm_package_consolePlugin_name}"
```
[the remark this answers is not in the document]
PeterYurkovich: fixed in latest commit to have npm_package_consolePlugin_name pull from CONSOLE_PLUGIN_NAME

PR #1089, opened 2026-07-23
OU-1389: Style Guide

PeterYurkovich:
/hold

Point of discussion for a potential style guide. All items should be enforced by lint rules where possible

### DIFF
--- a/docs/STYLEGUIDE.md
+++ b/docs/STYLEGUIDE.md
@@ -22,3 +22,38 @@ This document defines the monitoring-plugin repository style guides. It is a liv
 
 Existing files that predate this styleguide are not required to be renamed immediately. New files and renamed files must conform. Existing
 non-conforming files will be migrated incrementally over time without disrupting CI.
+
+New rules:
+
+### 1. Use function for component definition
+
+The displayName of a react component is often used in devtools and the like. When defining a component as a `function` it is able to use the function name, but when defining a function as an `arrow function` you need to add an additional `.displayName` to the variable.
+
+Needing to remember to add a duplicate `.displayName` to an `arrow function` component is extra mental overhead, so we should use `function` definitions instead.
+
+### 2. No default exports
+
+Do not allow default exports from files, instead require each function, variable, etc. to be exported manually
+
+### 3. Require Page files to only export react components ending with Page
+
+With our component structure and the need to export page modules in package.json and use their exported modules in `.patch.json` we should have consistent module exports as well as file names
+
+### 4. Avoid Higher-order Components
+
+Avoid usage of higher order components like `withFallback` and instead prefer manually wrapping components and context/hooks to deliver data
+
+### 5. React Prop names
+
+camelCase for props unless the prop is a React Component in which case it should be PascalCase
+
+### 6. `react/self-closing-comp`
+
+Always self close tags have no children
+
+Others for consideration:
+
+- Barrel Files
+- export all things at bottom of file (easy to find what each file exports)
+- Require commit signing
+- Require conventional commits https://www.conventionalcommits.org/en/v1.0.0/


docs/STYLEGUIDE.md
```diff
@@ -22,3 +22,38 @@ This document defines the monitoring-plugin repository style guides. It is a liv
...
+
+camelCase for props unless the prop is a React Component in which case it should be PascalCase
+
+### 6. `react/self-closing-comp`
+
+Always self close tags have no children
+
+Others for consideration:
+
+- Barrel Files
```
PeterYurkovich, line 56: I am personally not a fan of barrel files as they add a layer of indirection and don't add any functional change to the codebase. I would prefer either using alias's or just having the full path

docs/STYLEGUIDE.md
```diff
@@ -22,3 +22,38 @@ This document defines the monitoring-plugin repository style guides. It is a liv
...
+- export all things at bottom of file (easy to find what each file exports)
+- Require commit signing
+- Require conventional commits https://www.conventionalcommits.org/en/v1.0.0/
```
jgbernalp, line 59: +1

docs/STYLEGUIDE.md
jgbernalp: fair enough, I think they are also more useful for libraries rather than a consumer app

docs/STYLEGUIDE.md
```diff
@@ -22,3 +22,38 @@ This document defines the monitoring-plugin repository style guides. It is a liv
...
+camelCase for props unless the prop is a React Component in which case it should be PascalCase
+
+### 6. `react/self-closing-comp`
+
+Always self close tags have no children
+
+Others for consideration:
+
+- Barrel Files
+- export all things at bottom of file (easy to find what each file exports)
```
jgbernalp, line 57: I don't have a strong opinion about it, but I think IDEs can resolve where the export is so it might be just an overhead

docs/STYLEGUIDE.md
```diff
@@ -22,3 +22,38 @@ This document defines the monitoring-plugin repository style guides. It is a liv
...
+
+Avoid usage of higher order components like `withFallback` and instead prefer manually wrapping components and context/hooks to deliver data
+
+### 5. React Prop names
+
+camelCase for props unless the prop is a React Component in which case it should be PascalCase
+
+### 6. `react/self-closing-comp`
+
+Always self close tags have no children
```
jgbernalp, line 52:
```suggestion

### 6. `react/self-closing-comp`

Always self close tags have no children

### 7. co-locate test files

test files must have the `.spec.ts` suffix and be co-located with the file they are testing
```

docs/STYLEGUIDE.md
PeterYurkovich: Fair enough, I had just seen the pattern online and thought it was interesting to consider

docs/STYLEGUIDE.md
PeterYurkovich: Agree, maybe we should note this as being for unit and component tests

As I work through the dynamic test running and discuss more with @etmurasaki we can define the e2e tests placement later (maybe `/cypress/features/x` or `/src/features/x/tests`)

docs/STYLEGUIDE.md
```diff
@@ -22,3 +22,38 @@ This document defines the monitoring-plugin repository style guides. It is a liv
...
+
+Others for consideration:
+
```
PeterYurkovich, line 55: separate visualization from data, using hooks or wrapper components around visualization components

PeterYurkovich, line 55: When using Patternfly components consider creating "wrapper" components that our codebase will use which enable easier backporting by maintaining stable UI structure across versions

docs/STYLEGUIDE.md
zhuje: +1

PR #1113, opened 2026-08-05
NO-JIRA: feat: add renovate configuration

jgbernalp:
## Summary by CodeRabbit

* **Chores**
  * Added automated dependency update management with scheduled checks and controlled pull request limits.
  * Configured dependency version pinning, release-age safeguards, and grouped PatternFly updates.
  * Added vulnerability alert labeling and standardized security update commit messages.
  * Excluded selected dependencies and front-end packages from automated updates.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,60 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "schedule": ["* 6 * * *"],
+  "timezone": "UTC",
+  "prConcurrentLimit": 10,
+  "rangeStrategy": "pin",
+  "includePaths": ["go.mod", "go.sum", "web/**"],
+  "ignoreDeps": ["eslint", "eslint-config-base"],
+  "lockFileMaintenance": {
+    "enabled": false
+  },
+  "vulnerabilityAlerts": {
+    "labels": ["security"],
+    "commitMessagePrefix": "NO-JIRA: chore(security):"
+  },
+  "packageRules": [
+    {
+      "matchDepTypes": ["peerDependencies"],
+      "rangeStrategy": "auto"
+    },
+    {
+      "matchUpdateTypes": ["minor", "patch"],
+      "minimumReleaseAge": "14 days"
+    },
+    {
+      "matchFileNames": ["web/**"],
+      "matchPackageNames": [
+        "react",
+        "react-dom",
+        "react-i18next",
+        "react-redux",
+        "react-router",
+        "redux",
+        "redux-thunk",
+        "@perses-dev/*",
+        "@codemirror/*",
+        "@emotion/*",
+        "@hookform/resolvers",
+        "@lezer/highlight",
+        "@mui/material"
+      ],
+      "enabled": false
+    },
+    {
+      "matchFileNames": ["web/**"],
+      "matchPackageNames": [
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -0,0 +1,55 @@
...
+  "timezone": "UTC",
+  "prConcurrentLimit": 10,
+  "rangeStrategy": "pin",
+  "includePaths": ["go.mod", "go.sum", "web/**"],
+  "lockFileMaintenance": {
+    "enabled": false
+  },
+  "vulnerabilityAlerts": {
+    "labels": ["security"],
+    "commitMessageSuffix": "security:"
```
PeterYurkovich: This is the suffix, but I think with the `:` you are expecting the commit to look like `security: upgrade xyz` when its currently configured like `upgrade xyz security:`. I think based on conventional commit format it should be `chore(security):` as a `commitMessagePrefix that you are looking for

renovate.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "packageRules": [
+    {
+      "matchUpdateTypes": ["major", "minor", "patch"],
```
PeterYurkovich: I'm not super familiar with renovate, but will this cause renovate to create PR's to update all of our dependencies constantly even for breaking changes in major and minor versions?

renovate.json
```diff
@@ -0,0 +1,55 @@
...
+      "minimumReleaseAge": "14 days"
+    },
+    {
+      "matchFileNames": ["web/**"],
```
PeterYurkovich, line 26: We can just use [ignoreDeps](https://docs.renovatebot.com/configuration-options/#ignoredeps) instead of creating a group that only turns them off

renovate.json
```diff
@@ -0,0 +1,55 @@
...
+      "matchUpdateTypes": ["major", "minor"],
+      "enabled": false
+    },
+    {
+      "matchFileNames": ["web/**"],
+      "groupName": "patternfly",
+      "matchPackageNames": ["@patternfly/*"]
+    }
+  ]
+}
```
PeterYurkovich, line 60: We should probably just enable [semantic commits](https://docs.renovatebot.com/configuration-options/#semanticcommits).

Thinking about it more, it may be better to not have semantic commits for the autogenerated PR's and have the commit's start with `NO-JIRA:` instead so that they are able to be merged without us needing to update the title every time (or maybe this isn't necessary). https://docs.renovatebot.com/configuration-options/#prtitle

PeterYurkovich (commented):
Perhaps we can look at this as a follow up, but renovate does allow grouping updates, so we may be able to look into that to not be overwelmed

renovate.json
jgbernalp: I added both for now to align with our current and future commit message structure

renovate.json
jgbernalp: I'd like to keep it this way as the ignoreDeps is an exact match, and other ignored deps use pattern matching. I could split it but then there will be 2 places to think about.

renovate.json
jgbernalp: changed to `['minor', 'patch']`

jgbernalp:
> Perhaps we can look at this as a follow up, but renovate does allow grouping updates, so we may be able to look into that to not be overwelmed

yeah, I added some basic grouping but wanted to check after running it which others make sense.

PR #1123, opened 2026-08-12
OU-1486: [release-coo-ocp-4.19] refactor: remove i18n script in favor of env var configuration

jgbernalp:
- use webpack instead of i18n script
- remove obsolete Dockerfiles and i18n script


scripts/start-console.sh
```diff
@@ -4,9 +4,9 @@ set -euo pipefail
 
 CONSOLE_IMAGE=${CONSOLE_IMAGE:="quay.io/openshift/origin-console:latest"}
 CONSOLE_PORT=${CONSOLE_PORT:=9000}
-PLUGIN_PORT=${PLUGIN_PORT:=9001}
+PLUGIN_PORT=${PLUGIN_PORT:=9443}
```
PeterYurkovich, line 7: We need to update the readme to note that the backend needs to be running if we change the port here to use the backend port

jgbernalp: updated

PR #1124, opened 2026-08-12
OCPBUGS-99942: Fix missing query params in Prometheus API requests

gauron99:
## Claude analysis
Replace params.size > 0 with params.toString() check in buildPrometheusUrl. URLSearchParams.size is
unsupported in older browsers (pre-Chrome 113, pre-Firefox 112, pre-Safari 17), causing query parameters
to be silently dropped and Thanos to return 400.

https://caniuse.com/mdn-api_urlsearchparams_size

Affected versions: OCP 4.19 through current.
- 4.19-4.22: web/src/components/utils.ts
- 4.23+: web/src/shared/utils/utils.ts
- 4.18 and earlier: not affected.

## Verification

Headless Chrome test with `URLSearchParams.prototype.size` overridden to `undefined`:
- Old code (`params.size > 0`): returns empty string, query params dropped
- New code (`params.toString()`): returns `?query=up&start=123&end=456&step=30`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL generation for Prometheus queries, including cleaner handling when no search parameters are present.

### DIFF
--- a/web/src/shared/utils/utils.ts
+++ b/web/src/shared/utils/utils.ts
@@ -269,9 +269,8 @@ export const buildPrometheusUrl = ({
   }
 
   const params = getSearchParams(prometheusUrlProps);
-  return `${basePath}/${prometheusUrlProps.endpoint}${
-    params.size > 0 ? '?' + params.toString() : ''
-  }`;
+  const queryString = params.toString();
+  return `${basePath}/${prometheusUrlProps.endpoint}${queryString ? '?' + queryString : ''}`;
 };
 
 type PrometheusURLProps = {


PeterYurkovich:
@gauron99 Could you please define where the need to support browsers that are more than 3 years old is coming from? If it is a requirement for UI teams, then our team will need to add other polyfills and testing to ensure compliance

gauron99:
Hi, we are waiting to hear back about the CU browser version. I just prepped this PR for the case as it fixed my issue. @PeterYurkovich

gauron99:
closing the ticket as its still unclear what the JIRA resolution looks like, waiting on customer. keeping the reference to this PR as "immediate quick fix for the JIRA raised"

PR #1142, opened 2026-08-14
[release-4.22] OCPBUGS-98787: fix for CVE-2026-48801

rhdmalone:
[release-4.22] OCPBUGS-98787: fix for CVE-2026-48801

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -18826,12 +18826,22 @@
       "license": "MIT"
     },
     "node_modules/linkify-it": {
-      "version": "2.2.0",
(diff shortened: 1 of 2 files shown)


jgbernalp:
@rhdmalone this patch might introduce undesired behavior as is bumping the version at lease 4 versions ahead. The root cause is that the library is not maintained anymore. I created this PR to replace it with a maintained library. We should apply the same patch to older versions to resolve this CVE. https://github.com/openshift/monitoring-plugin/pull/1144

rhdmalone:
thanks @jgbernalp I will adopt the approach you took

PR #1148, opened 2026-08-18
NO-JIRA: chore: organize dev dependencies and avoid cypress binary install

jgbernalp:
- Avoid cypress binary installation in Dockerfiles, except Dockerfile.test
- organize dev deps for better CVE reporting

## Summary by CodeRabbit

- **Build Improvements**
  - Streamlined application and development container builds by skipping unnecessary Cypress binary downloads.
  - Improved test image creation by disabling automated Git hooks during builds.

- **Maintenance**
  - Reorganized frontend packages so build-time tools are separated from runtime dependencies.
  - Reduced unnecessary installation overhead and improved consistency across development, testing, and production environments.


PeterYurkovich:
Cool find on the cypress not installing env. We could probably do something similar for our build binaries as well, or at least only install the platform we need, although it's probably best to wait until we move to rspack to look into that.

I've set myself a task for today to split up the points we discussed in the testing meeting, split the styleguide PR into jira issues, and point the existing tech debt stories. I'll make sure to add in the rspack upgrade as well

PeterYurkovich:
/label qe-approved
QE performed by passing tests

PR #1171, opened 2026-08-24
CNV-87531: k8s: add Prometheus query layer and GET /alerts

sradco:
Add Prometheus, Thanos, and Alertmanager
query support with GET /api/v1/alerting/
alerts endpoint including alert component
matching and alerting health status.

## Summary by CodeRabbit

- **New Features**
  - Added the `GET /api/v1/alerting/alerts` endpoint for retrieving alerts.
  - Supports filtering by alert state, labels, and namespace-scoped results.
  - Enriches alerts with rule details, source, severity, component, and layer information.
  - Added alerting health reporting for monitoring routes and workload monitoring.
- **Bug Fixes**
  - Added validation and clear warnings for invalid filters, unavailable monitoring routes, and configuration issues.
  - Improved alert and rule retrieval across platform and workload monitoring sources.

### DIFF
--- a/internal/managementrouter/alerts_get.go
+++ b/internal/managementrouter/alerts_get.go
@@ -1,0 +1,109 @@
+package managementrouter
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/openshift/monitoring-plugin/pkg/k8s"
+)
+
+type GetAlertsResponse struct {
+	Data     GetAlertsResponseData `json:"data"`
+	Warnings []string              `json:"warnings,omitempty"`
+}
+
+type GetAlertsResponseData struct {
+	Alerts []k8s.PrometheusAlert `json:"alerts"`
+}
+
+func (hr *httpRouter) GetAlerts(w http.ResponseWriter, req *http.Request) {
+	state, labels, err := parseStateAndLabels(req.URL.Query())
+	if err != nil {
+		writeError(w, http.StatusBadRequest, err.Error())
+		return
+	}
+	ctx := req.Context()
+
+	alerts, err := hr.managementClient.GetAlerts(ctx, k8s.GetAlertsRequest{
+		Labels: labels,
+		State:  state,
+	})
+	if err != nil {
+		handleError(w, err)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Cache-Control", "no-store")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(GetAlertsResponse{
+		Data: GetAlertsResponseData{
+			Alerts: alerts,
+		},
+		Warnings: hr.alertWarnings(ctx),
+	}); err != nil {
+		log.WithError(err).Warn("failed to encode alerts response")
+	}
+}
+
+func (hr *httpRouter) alertWarnings(ctx context.Context) []string {
+	health, ok := hr.alertingHealth(ctx)
+	if !ok {
+		return nil
+	}
+
+	warnings := []string{}
+	if health.UserWorkloadEnabled && health.UserWorkload != nil {
+		warnings = append(warnings, buildRouteWarnings(health.UserWorkload.Prometheus, k8s.UserWorkloadRouteName, "user workload Prometheus")...)
+		warnings = append(warnings, buildRouteWarnings(health.UserWorkload.Alertmanager, k8s.UserWorkloadAlertmanagerRouteName, "user workload Alertmanager")...)
+	}
+
+	return warnings
+}
+
+//nolint:unused // used by the rules listing handler in a subsequent branch
+func (hr *httpRouter) rulesWarnings(ctx context.Context) []string {
+	health, ok := hr.alertingHealth(ctx)
+	if !ok {
+		return nil
+	}
+
+	if health.UserWorkloadEnabled && health.UserWorkload != nil {
+		return buildRouteWarnings(health.UserWorkload.Prometheus, k8s.UserWorkloadRouteName, "user workload Prometheus")
+	}
+
+	return nil
+}
+
+func (hr *httpRouter) alertingHealth(ctx context.Context) (k8s.AlertingHealth, bool) {
+	if hr.managementClient == nil {
+		return k8s.AlertingHealth{}, false
+	}
+
+	health, err := hr.managementClient.GetAlertingHealth(ctx)
+	if err != nil {
+		log.WithError(err).Warn("alerting health unavailable")
+		return k8s.AlertingHealth{}, false
+	}
+
+	return health, true
+}
+
+func buildRouteWarnings(route k8s.AlertingRouteHealth, expectedName string, friendlyName string) []string {
+	if route.Name != "" && route.Name != expectedName {
+		return nil
+	}
+	if route.FallbackReachable {
+		return nil
+	}
+
+	switch route.Status {
+	case k8s.RouteNotFound:
+		return []string{friendlyName + " route is missing"}
+	case k8s.RouteUnreachable:
+		return []string{friendlyName + " route is unreachable"}
+	default:
+		return nil
+	}
+}
--- a/internal/managementrouter/alerts_get.go
+++ b/internal/managementrouter/alerts_get.go
@@ -1,0 +1,96 @@
+package managementrouter
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/openshift/monitoring-plugin/pkg/k8s"
+)
+
+type GetAlertsResponse struct {
+	Data     GetAlertsResponseData `json:"data"`
+	Warnings []string              `json:"warnings,omitempty"`
+}
+
+type GetAlertsResponseData struct {
+	Alerts []k8s.PrometheusAlert `json:"alerts"`
+}
+
+func (hr *httpRouter) GetAlerts(w http.ResponseWriter, req *http.Request) {
+	state, labels, err := parseStateAndLabels(req.URL.Query())
+	if err != nil {
+		writeError(w, http.StatusBadRequest, err.Error())
+		return
+	}
+	ctx := req.Context()
+
+	alerts, warnings, err := hr.managementClient.EnrichAlerts(ctx, k8s.GetAlertsRequest{
+		Labels: labels,
+		State:  state,
+	})
+	if err != nil {
+		handleError(w, err)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Cache-Control", "no-store")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(GetAlertsResponse{
+		Data: GetAlertsResponseData{
+			Alerts: alerts,
+		},
+		Warnings: warnings,
+	}); err != nil {
+		log.WithError(err).Warn("failed to encode alerts response")
+	}
+}
+
+//nolint:unused // used by the rules listing handler in a subsequent branch
+func (hr *httpRouter) rulesWarnings(ctx context.Context) []string {
+	health, ok := hr.alertingHealth(ctx)
+	if !ok {
+		return nil
+	}
+
+	if health.UserWorkloadEnabled && health.UserWorkload != nil {
+		return buildRouteWarnings(health.UserWorkload.Prometheus, k8s.UserWorkloadRouteName, "user workload Prometheus")
+	}
+
+	return nil
+}
+
+//nolint:unused // called by rulesWarnings, used in a subsequent branch
+func (hr *httpRouter) alertingHealth(ctx context.Context) (k8s.AlertingHealth, bool) {
+	if hr.managementClient == nil {
+		return k8s.AlertingHealth{}, false
+	}
+
+	health, err := hr.managementClient.GetAlertingHealth(ctx)
+	if err != nil {
+		log.WithError(err).Warn("alerting health unavailable")
+		return k8s.AlertingHealth{}, false
+	}
+
+	return health, true
+}
+
+//nolint:unused // called by rulesWarnings, used in a subsequent branch
+func buildRouteWarnings(route k8s.AlertingRouteHealth, expectedName string, friendlyName string) []string {
+	if route.Name != "" && route.Name != expectedName {
+		return nil
+	}
+	if route.FallbackReachable {
+		return nil
+	}
+
+	switch route.Status {
+	case k8s.RouteNotFound:
+		return []string{friendlyName + " route is missing"}
+	case k8s.RouteUnreachable:
+		return []string{friendlyName + " route is unreachable"}
+	default:
+		return nil
+	}
+}
--- a/internal/managementrouter/query_filters.go
+++ b/internal/managementrouter/query_filters.go
@@ -1,0 +1,35 @@
+package managementrouter
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+var validStates = map[string]bool{
+	"":         true,
+	"pending":  true,
+	"firing":   true,
+	"silenced": true,
+}
+
+// parseStateAndLabels returns the optional state filter and label matches.
+// Any query param other than "state" is treated as a label match.
+// Returns an error if the state value is not one of the known states.
+func parseStateAndLabels(q url.Values) (string, map[string]string, error) {
+	state := strings.ToLower(strings.TrimSpace(q.Get("state")))
+	if !validStates[state] {
+		return "", nil, fmt.Errorf("invalid state filter %q: must be one of pending, firing, silenced", q.Get("state"))
+	}
+
+	labels := make(map[string]string)
+	for key, vals := range q {
+		if key == "state" {
+			continue
+		}
+		if len(vals) > 0 && strings.TrimSpace(vals[0]) != "" {
+			labels[strings.TrimSpace(key)] = strings.TrimSpace(vals[0])
+		}
+	}
+	return state, labels, nil
+}
--- a/internal/managementrouter/query_filters.go
+++ b/internal/managementrouter/query_filters.go
@@ -1,0 +1,38 @@
+package managementrouter
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+var validStates = map[string]bool{
+	"pending":  true,
+	"firing":   true,
+	"silenced": true,
+}
+
+// parseStateAndLabels returns the optional state filter and label matches.
+// Any query param other than "state" is treated as a label match.
+// Returns an error if the state value is not one of the known states.
+func parseStateAndLabels(q url.Values) (string, map[string]string, error) {
+	state := strings.ToLower(strings.TrimSpace(q.Get("state")))
+	if state != "" && !validStates[state] {
+		return "", nil, fmt.Errorf("invalid state filter %q: must be one of pending, firing, silenced", state)
+	}
+
+	labels := make(map[string]string)
+	for key, vals := range q {
+		if key == "state" {
+			continue
+		}
+		if len(vals) == 0 || strings.TrimSpace(vals[0]) == "" {
+			continue
+		}
+		if len(vals) > 1 {
+			return "", nil, fmt.Errorf("multiple values for label filter %q: only a single value is supported", key)
+		}
+		labels[strings.TrimSpace(key)] = strings.TrimSpace(vals[0])
+	}
+	return state, labels, nil
+}
--- a/pkg/alertcomponent/matcher.go
+++ b/pkg/alertcomponent/matcher.go
@@ -1,0 +1,381 @@
+package alertcomponent
+
+import (
+	"regexp"
+
+	"github.com/prometheus/common/model"
+
+	"github.com/openshift/monitoring-plugin/pkg/managementlabels"
+)
+
+const (
+	labelNamespace = "namespace"
+	labelSeverity  = "severity"
+)
+
+func ns(values ...string) LabelsMatcher {
+	return NewLabelsMatcher(labelNamespace, NewStringValuesMatcher(values...))
+}
+
+func alertNames(values ...string) LabelsMatcher {
+	return NewLabelsMatcher(managementlabels.AlertNameLabel, NewStringValuesMatcher(values...))
+}
+
+func regexAlertNames(regexes ...*regexp.Regexp) LabelsMatcher {
+	return NewLabelsMatcher(managementlabels.AlertNameLabel, NewRegexValuesMatcher(regexes...))
+}
+
+func labelValues(key string, values ...string) LabelsMatcher {
+	return NewLabelsMatcher(key, NewStringValuesMatcher(values...))
+}
+
+func comp(component string, ms ...LabelsMatcher) componentMatcher {
+	return componentMatcher{component: component, matchers: ms}
+}
+
+// LabelsMatcher represents a matcher definition for a set of labels.
+// It matches if all of the label matchers match the labels.
+type LabelsMatcher interface {
+	Matches(labels model.LabelSet) (match bool, keys []model.LabelName)
+	Equals(other LabelsMatcher) bool
+}
+
+func NewLabelsMatcher(key string, matcher ValueMatcher) LabelsMatcher {
+	return labelMatcher{key: key, matcher: matcher}
+}
+
+func NewStringValuesMatcher(keys ...string) ValueMatcher {
+	return stringMatcher(keys)
+}
+
+func NewRegexValuesMatcher(regexes ...*regexp.Regexp) ValueMatcher {
+	return regexpMatcher(regexes)
+}
+
+// labelMatcher represents a matcher definition for a label.
+type labelMatcher struct {
+	key     string
+	matcher ValueMatcher
+}
+
+// Matches implements the LabelsMatcher interface.
+func (l labelMatcher) Matches(labels model.LabelSet) (bool, []model.LabelName) {
+	if l.matcher.Matches(string(labels[model.LabelName(l.key)])) {
+		return true, []model.LabelName{model.LabelName(l.key)}
+	}
+	return false, nil
+}
+
+// Equals implements the LabelsMatcher interface.
+func (l labelMatcher) Equals(other LabelsMatcher) bool {
+	ol, ok := other.(labelMatcher)
+	if !ok {
+		return false
+	}
+	return l.key == ol.key && l.matcher.Equals(ol.matcher)
+}
+
+// ValueMatcher represents a matcher for a specific value.
+//
+// Multiple implementations are provided for different types of matchers.
+type ValueMatcher interface {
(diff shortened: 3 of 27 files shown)


internal/managementrouter/alerts_get.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Cache-Control", "no-store")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(GetAlertsResponse{
+		Data: GetAlertsResponseData{
+			Alerts: alerts,
+		},
+		Warnings: hr.alertWarnings(ctx),
```
PeterYurkovich: At it's root, alertWarnings is used to fetch the `/health` endpoint of platform and user workload prometheus's, then surface error in connecting to the `/health` endpoint to users. 

The purpose behind this is that within `hr.managementClient.GetAlerts` we don't want to immediately return an error if one of the endpoints fails to return correctly, so instead we `log.warn` and proceed forward with processing any return from the other.

I think this approach is overcomplicated and actually doesn't surface the users issues. Requests could be denied for reasons other than the `/health` endpoint being down which the users should know about. Really what we want is failures of each endpoint within `GetAlerts` and to send them back as warnings. So we should just create a slice of warnings in `GetAlerts` and return them back, then surface those warnings to the user (maybe after some formatting if need be)

internal/managementrouter/alerts_get.go
```diff
@@ -0,0 +1,109 @@
...
+type GetAlertsResponse struct {
+	Data     GetAlertsResponseData `json:"data"`
+	Warnings []string              `json:"warnings,omitempty"`
+}
+
+type GetAlertsResponseData struct {
+	Alerts []k8s.PrometheusAlert `json:"alerts"`
+}
+
+func (hr *httpRouter) GetAlerts(w http.ResponseWriter, req *http.Request) {
```
PeterYurkovich, line 20: With the layers of abstractions it can be difficult to follow the actual functions being used in this PR due to many functions sharing the same name. Since there are 3 separate `GetAlerts` functions you can't grep/search for the name and the abstaction through interfaces means you can't use LSP's actions to "go to definition" or "find all references". I know it makes the individual functions not as clear to their exact purpose but could we swap to different function names so that we can grep/search through the codebase and only see the exact function being searched for. Same for some other the functions like alertinghealth

internal/managementrouter/query_filters.go
```diff
@@ -0,0 +1,35 @@
+package managementrouter
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+var validStates = map[string]bool{
+	"":         true,
```
PeterYurkovich: Based on the error message in the !validStates below, empty string shouldn't be allowed correct?

internal/managementrouter/query_filters.go
```diff
@@ -0,0 +1,35 @@
...
+	"silenced": true,
+}
+
+// parseStateAndLabels returns the optional state filter and label matches.
+// Any query param other than "state" is treated as a label match.
+// Returns an error if the state value is not one of the known states.
+func parseStateAndLabels(q url.Values) (string, map[string]string, error) {
+	state := strings.ToLower(strings.TrimSpace(q.Get("state")))
+	if !validStates[state] {
+		return "", nil, fmt.Errorf("invalid state filter %q: must be one of pending, firing, silenced", q.Get("state"))
```
PeterYurkovich:
```suggestion
		return "", nil, fmt.Errorf("invalid state filter %q: must be one of pending, firing, silenced", state)
```

internal/managementrouter/query_filters.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	labels := make(map[string]string)
+	for key, vals := range q {
+		if key == "state" {
+			continue
+		}
+		if len(vals) > 0 && strings.TrimSpace(vals[0]) != "" {
+			labels[strings.TrimSpace(key)] = strings.TrimSpace(vals[0])
```
PeterYurkovich: Only the first label value is used, so we should either extend this to work with multiple label values or error if multiple values are present as it won't work as people expect

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
...
+	Equals(other ValueMatcher) bool
+}
+
+// stringMatcher is a matcher for a list of strings.
+//
+// It matches if the value is in the list of strings.
+type stringMatcher []string
+
+func (s stringMatcher) Matches(value string) bool {
+	for _, v := range s {
```
PeterYurkovich: can be a `slices.Contains` call

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
...
+		s1 = append(s1, re.String())
+	}
+	s2 := make([]string, 0, len(o))
+	for _, re := range o {
+		s2 = append(s2, re.String())
+	}
+	return equalsNoOrder(s1, s2)
+}
+
+func equalsNoOrder(a, b []string) bool {
```
PeterYurkovich, line 141: Why would we want to match on strings that are mixed up?

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
+package alertcomponent
```
PeterYurkovich, line 1: Could we add some test cases for this file?

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
...
+		comp("openshift-local-storage", ns("openshift-local-storage")),
+		comp("quay", labelValues("container", "quay-app", "quay-mirror", "quay-app-upgrade")),
+		comp("Argo", regexAlertNames(regexp.MustCompile("^Argo"))),
+	}
+)
+
+var cvoAlerts = []model.LabelValue{"ClusterOperatorDown", "ClusterOperatorDegraded"}
+
+func cvoAlertsMatcher(labels model.LabelSet) (layer, comp model.LabelValue, keys []model.LabelName) {
+	for _, v := range cvoAlerts {
```
PeterYurkovich: can be a silces.Contains

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	return "cluster", "OpenShift Virtualization Operator", []model.LabelName{
+		"kubernetes_operator_part_of",
+		"kubernetes_operator_component",
+		"operator_health_impact",
+	}
+}
+
+func computeMatcher(labels model.LabelSet) (layer, comp model.LabelValue, keys []model.LabelName) {
+	for _, nodeAlert := range nodeAlerts {
```
PeterYurkovich: can be a slices.contains

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
...
+
+func regexAlertNames(regexes ...*regexp.Regexp) LabelsMatcher {
+	return NewLabelsMatcher(managementlabels.AlertNameLabel, NewRegexValuesMatcher(regexes...))
+}
+
+func labelValues(key string, values ...string) LabelsMatcher {
+	return NewLabelsMatcher(key, NewStringValuesMatcher(values...))
+}
+
+func comp(component string, ms ...LabelsMatcher) componentMatcher {
```
PeterYurkovich: Could we make theses 5 functions names a bit more explicit, ie. namespaceMatcher. Reading the usage below is a bit confusing as its unclear what comp stands for (component, componentMatcher or compare)

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
...
+type componentMatcherFn func(labels model.LabelSet) (layer, comp model.LabelValue, keys []model.LabelName)
+
+func evalMatcherFns(fns []componentMatcherFn, labels model.LabelSet) (
+	layer, comp string, labelsSubset model.LabelSet) {
+	for _, fn := range fns {
+		if layer, comp, keys := fn(labels); layer != "" {
+			return string(layer), string(comp), getLabelsSubset(labels, keys...)
+		}
+	}
+	return "Others", "Others", getLabelsSubset(labels)
```
PeterYurkovich, line 195: `alert-rule-classification.md` needs the `2) Fallback:` section updated to state that fallback of layer is "Others"

pkg/management/get_alerts.go
```diff
@@ -0,0 +1,308 @@
...
+			lbls["namespace"] = model.LabelValue(ns)
+		}
+	}
+	if rule.Alert != "" {
+		lbls[model.LabelName(managementlabels.AlertNameLabel)] = model.LabelValue(rule.Alert)
+	}
+
+	layer, component := alertcomponent.DetermineComponent(lbls)
+	if component == "" || component == "Others" {
+		component = "other"
```
PeterYurkovich: We should keep the component and layer string consistent as `"Others"` while being used interally

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
...
+		comp("cloud-credential", ns("openshift-cloud-credential-operator")),
+		comp("cluster-api", ns("openshift-cluster-api", "openshift-cluster-api-operator")),
+		comp("config-operator", ns("openshift-config-operator")),
+		comp("kube-storage-version-migrator", ns("openshift-kube-storage-version-migrator", "openshift-kube-storage-version-migrator-operator")),
+		comp("image-registry", ns("openshift-image-registry", "openshift-image-registry-operator")),
+		comp("ingress", ns("openshift-ingress", "openshift-route-controller-manager", "openshift-ingress-canary", "openshift-ingress-operator")),
+		comp("console", ns("openshift-console", "openshift-console-operator")),
+		comp("insights", ns("openshift-insights", "openshift-insights-operator")),
+		comp("machine-api", ns("openshift-machine-api", "openshift-machine-api-operator")),
+		comp("monitoring", ns("openshift-monitoring", "openshift-monitoring-operator")),
```
PeterYurkovich:
```suggestion
		comp("monitoring", ns(ClusterMonitoringNamespace, "openshift-monitoring-operator")),
```

pkg/alertcomponent/matcher.go
```diff
@@ -0,0 +1,381 @@
...
+		comp("storage", ns("openshift-storage", "openshift-cluster-csi-drivers", "openshift-cluster-storage-operator", "openshift-storage-operator")),
+		comp("vertical-pod-autoscaler", ns("openshift-vertical-pod-autoscaler", "openshift-vertical-pod-autoscaler-operator")),
+		comp("marketplace", ns("openshift-marketplace", "openshift-marketplace-operator")),
+	}
+
+	workloadMatchers = []componentMatcher{
+		comp("openshift-compliance", ns("openshift-compliance")),
+		comp("openshift-file-integrity", ns("openshift-file-integrity")),
+		comp("openshift-logging", ns("openshift-logging")),
+		comp("openshift-user-workload-monitoring", ns("openshift-user-workload-monitoring")),
```
PeterYurkovich:
```suggestion
		comp("openshift-user-workload-monitoring", ns(UserWorkloadMonitoringNamespace)),
```

pkg/k8s/prometheus_alerts.go
```diff
@@ -0,0 +1,949 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+)
+
+var prometheusLog = logrus.WithField("module", "k8s-prometheus")
+
+const (
+	namespaceCacheTTL      = 30 * time.Second
+	serviceHealthTimeout   = 5 * time.Second
+	serviceRequestTimeout  = 10 * time.Second
```
PeterYurkovich, line 31: I'll let @simonpasquier and @jgbernalp chime in here, but one thing that has constantly come up for projects we have shipped is a customer desire to tune constants, cache ttls and the like. Should we expose these through parameters or in the config file?

pkg/k8s/prometheus_alerts.go
```diff
@@ -0,0 +1,949 @@
...
+
+// GetAlertsRequest holds parameters for filtering alerts
+type GetAlertsRequest struct {
+	// Labels filters alerts by labels
+	Labels map[string]string
+	// State filters alerts by state: "firing", "pending", "silenced", or "" for all states
+	State string
+}
+
+type PrometheusAlert struct {
```
PeterYurkovich, line 88: Could be a bad idea, but since this is supposed to be a superset of the prometheus alert should we add this as a wrapper around the promethues common Alert ([link](https://pkg.go.dev/github.com/prometheus/common@v0.70.1/model#Alert))?

pkg/k8s/prometheus_alerts.go
```diff
@@ -0,0 +1,949 @@
...
+		if !labelsMatch(&req, &a) {
+			continue
+		}
+
+		out = append(out, a)
+	}
+	return out, nil
+}
+
+func matchesAlertState(requestedState string, alertState string) bool {
```
PeterYurkovich: Should probably move this to the matchers helper file for consistency

pkg/k8s/prometheus_alerts.go
```diff
@@ -0,0 +1,949 @@
...
+	amAlerts, amErr := pa.getAlertmanagerAlerts(ctx, namespace, amRouteName, source)
+	promAlerts, promErr := pa.getAlertsViaProxy(ctx, namespace, promRouteName, source)
+
+	if amErr == nil {
+		pending := filterAlertsByState(promAlerts, "pending")
+		return append(amAlerts, pending...), nil
+	}
+
+	if promErr != nil {
+		return nil, promErr
```
PeterYurkovich: amError gets eaten if both services error

pkg/management/get_alerts.go
```diff
@@ -0,0 +1,308 @@
...
+		if alert.Labels[k8s.AlertSourceLabel] != k8s.AlertSourceUser {
+			relabels, keep := relabel.Process(labels.FromMap(alert.Labels), configs...)
+			if !keep {
+				continue
+			}
+			alert.Labels = relabels.Map()
+		}
+
+		// Add calculated rule ID and source when not present (labels enrichment)
+		c.setRuleIDAndSourceIfMissing(ctx, &alert, rules)
```
PeterYurkovich, line 48: What happens if the computed rule ID and the one present on the rule is different?

pkg/k8s/prometheus_alerts.go
sradco: I checked the codebase, all other timeouts/TTLs (proxy, health, init) are hardcoded constants.
Only the top-level server timeout is configurable via plugin config. 
I think we should keep these as sensible defaults for now and add configurability if a real need arises.

internal/managementrouter/alerts_get.go
sradco: Done. 
I redesigned warning propagation: 
`FetchAlerts` now collects non-fatal errors as `[]string` warnings at the point of failure and propagates them through `EnrichAlerts` to the handler. 
The separate `alertWarnings` + health-check code path is removed. 
This avoids the stale-state issue where health could change between fetch and check.

internal/managementrouter/alerts_get.go
sradco: Done. 
I renamed across layers: 
- `FetchAlerts` - k8s  raw data retrieval
-  `EnrichAlerts` - management -  adds classification
-  `GetAlerts` - HTTP handler - serves request.
Each name is now unique and greppable.

internal/managementrouter/query_filters.go
sradco: Correct. 
I removed `"": true` from the map.
Empty state now passes through (matches all alerts) but any non-empty invalid state returns an error. 
The validation is `if state != "" && !validStates[state]`.

internal/managementrouter/query_filters.go
sradco: I applied your suggestion.
The error message now uses the normalized state value.

internal/managementrouter/query_filters.go
sradco: I extend this to work with multiple label values

pkg/alertcomponent/matcher.go
sradco: I applied `slices.Contains` in all three locations: `stringMatcher.Matches`, `cvoAlertsMatcher`, and `computeMatcher`.

pkg/alertcomponent/matcher.go
sradco: `equalsNoOrder` is used by `ValueMatcher.Equals` to compare two matchers for logical equivalence regardless of the order values were specified.
For example, `NewStringValuesMatcher("a", "b")` and `NewStringValuesMatcher("b", "a")` represent the same match set.
It's not about mixing user-facing strings - it's internal matcher comparison for deduplication/equality checks.

pkg/alertcomponent/matcher.go
sradco: Done. Added `matcher_test.go`.

pkg/alertcomponent/matcher.go
sradco: Done. Renamed all five helper functions:

`ns` → `namespaceMatcher`
`comp` → `componentRule`
`alertNames` → `alertNameMatcher`
`regexAlertNames` → `regexAlertNameMatcher`
`labelValues` → `labelValueMatcher`

pkg/alertcomponent/matcher.go
sradco: Done. 
I updated `docs/alert-rule-classification.md` fallback section: `component = other` to  `component = "Others"`.

pkg/management/get_alerts.go
sradco: Done. 
I changed both `classifyFromRule` and `classifyFromAlertLabels` to use `"Others"` for the fallback component. Updated tests to match.

pkg/alertcomponent/matcher.go
sradco: Done.

pkg/alertcomponent/matcher.go
sradco: Done.

pkg/k8s/prometheus_alerts.go
sradco: `model.Alert` has `Labels`, `Annotations`, `StartsAt`, `EndsAt`, `GeneratorURL` - a subset of what `PrometheusAlert` carries. 
The gap is that `PrometheusAlert` also has `State`, `ActiveAt`, `AlertRuleId`, `AlertComponent`, `AlertLayer`, and routing labels (`openshift_io_alert_source`, etc.) that don't exist in the common model. 
Embedding `model.Alert` would save 5 fields but add a dependency on the Prometheus model in our API response type. 
I'd defer this to a follow-up since it's a structural change that touches the JSON serialization contract. 
Happy to discuss if you feel strongly.

pkg/k8s/prometheus_alerts.go
sradco: Done

pkg/k8s/prometheus_alerts.go
sradco: Good catch! 
Fixed.

pkg/management/get_alerts.go
sradco: `setRuleIDAndSourceIfMissing` only sets the rule ID when the alert doesn't already have one (check at the top: `if alert.AlertRuleId != "" { return }`).
So if a rule ID is already present, it's preserved.
The computed ID is a fallback for alerts that arrive without one. 
If you think we should also validate consistency when both exist, we can add that as a follow-up.

internal/managementrouter/alerts_get.go
```diff
@@ -0,0 +1,94 @@
...
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Cache-Control", "no-store")
+	w.WriteHeader(http.StatusOK)
+	if err := json.NewEncoder(w).Encode(GetAlertsResponse{
+		Data: GetAlertsResponseData{
+			Alerts: alerts,
+		},
+		Warnings: warnings,
+	}); err != nil {
```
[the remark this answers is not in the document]
sradco: Done.

internal/managementrouter/query_filters.go
```diff
@@ -0,0 +1,38 @@
...
+	"pending":  true,
+	"firing":   true,
+	"silenced": true,
+}
+
+// parseStateAndLabels returns the optional state filter and label matches.
+// Any query param other than "state" is treated as a label match.
+// Returns an error if the state value is not one of the known states.
+func parseStateAndLabels(q url.Values) (string, map[string]string, error) {
+	state := strings.ToLower(strings.TrimSpace(q.Get("state")))
```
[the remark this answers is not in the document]
sradco: Done.

pkg/k8s/prometheus_alerts_test.go
```diff
@@ -0,0 +1,331 @@
...
+import (
+	"encoding/json"
+	"testing"
+	"time"
+)
+
+// --- parseAlertmanagerResponse ---
+
+func TestParseAlertmanagerResponse_ValidAlerts(t *testing.T) {
+	raw, _ := json.Marshal([]alertmanagerAlert{
```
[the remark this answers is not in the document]
sradco: Done.

pkg/k8s/prometheus_alerts.go
```diff
@@ -0,0 +1,930 @@
...
+
+	if promErr != nil {
+		return nil, fmt.Errorf("alertmanager: %w; prometheus: %w", amErr, promErr)
+	}
+
+	return promAlerts, nil
+}
+
+func (pa *prometheusAlerts) getUserWorkloadAlerts(ctx context.Context, req GetAlertsRequest) ([]PrometheusAlert, error) {
+	if shouldPreferUserAlertmanager(req.State) {
```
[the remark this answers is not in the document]
sradco: Done.

pkg/k8s/prometheus_alerts.go
```diff
@@ -0,0 +1,930 @@
...
+func fanOutThanosTenancy[T any](namespaces []string, fetch func(string) ([]T, error)) ([]T, error) {
+	var out []T
+	var lastErr error
+	for _, namespace := range namespaces {
+		results, err := fetch(namespace)
+		if err != nil {
+			lastErr = err
+			continue
+		}
+		out = append(out, results...)
```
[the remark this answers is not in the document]
sradco: Done.

pkg/k8s/prometheus_alerts.go
```diff
@@ -0,0 +1,930 @@
...
+
+	return strings.TrimSpace(string(tokenBytes)), nil
+}
+
+func (pa *prometheusAlerts) performRequest(client *http.Client, req *http.Request) ([]byte, error) {
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("execute request: %w", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
```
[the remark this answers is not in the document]
sradco: Done.

test/e2e/framework/framework.go
```diff
@@ -332,6 +332,64 @@ func (f *Framework) CreateScopedUser(ctx context.Context, name, namespace, apiGr
...
 
+// CreateUserWithClusterRole creates a ServiceAccount in the given namespace and
+// binds an existing ClusterRole to it via a namespaced RoleBinding. Use this
+// for OpenShift built-in roles such as monitoring-rules-view (Thanos tenancy
+// on port 9093). API calls are retried to tolerate transient failures.
+func (f *Framework) CreateUserWithClusterRole(ctx context.Context, name, namespace, clusterRoleName string) (*ScopedUser, error) {
+	rollback := func() {
+		_ = f.Clientset.RbacV1().RoleBindings(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+		_ = f.Clientset.CoreV1().ServiceAccounts(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	}
```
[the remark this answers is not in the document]
sradco: Done.

test/e2e/get_alerts_test.go
```diff
@@ -0,0 +1,317 @@
...
+	defer func() {
+		if err := userB.Cleanup(); err != nil {
+			t.Logf("cleanup user B failed: %v", err)
+		}
+	}()
+
+	alertName := "E2ERBACGetAlertsTest"
+	if err := createFiringPrometheusRule(ctx, f, nsY, "e2e-rbac-get-alerts-rule", alertName); err != nil {
+		t.Fatalf("Failed to create PrometheusRule: %v", err)
+	}
```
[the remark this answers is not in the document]
sradco: Done.

test/e2e/framework/framework.go
```diff
@@ -332,6 +332,64 @@ func (f *Framework) CreateScopedUser(ctx context.Context, name, namespace, apiGr
...
+
+	token, err := f.requestServiceAccountToken(ctx, namespace, name)
+	if err != nil {
+		rollback()
+		return nil, err
+	}
+
+	return &ScopedUser{Token: token, Cleanup: func() error { rollback(); return nil }}, nil
+}
```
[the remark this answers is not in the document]
sradco: Added `framework_test.go` with 6 tests for retry and Poll helpers (the newly-added retry function was previously missing). 
Full unit testing of `CreateUserWithClusterRole` requires refactoring Framework to accept `kubernetes.Interface` instead of `*kubernetes.Clientset` - deferring that structural change.

pkg/management/get_alerts.go
PeterYurkovich: Thought about it some more and I think it's best to leave as is to allow for a single ID for the lifecycle of the alert

pkg/k8s/prometheus_alerts.go
PeterYurkovich: Lets keep as is

pkg/alertcomponent/matcher.go
PeterYurkovich: Alright, I understand it now. While I think the rest of the matchers functions can be pretty self explanatory the logic in this function is kinda hard to parse and hold in your head. Would you mind making the variables more explicit. Something like 

```go
func equalsNoOrder(matcherSliceA, matcherSliceB []string) bool {
	if len(matcherSliceA) != len(matcherSliceB) {
		return false
	}

	seen := make(map[string]int, len(matcherSliceA))
	for _, matcher := range matcherSliceA {
		seen[matcher]++
	}
	// Since we validated the slices length's are the same above, any difference between  
	// the slices will always result in there being less of a particular matcher in matcherSliceB
	for _, matcher := range matcherSliceB {
		if seen[matcher] == 0 {
			return false
		}
		seen[matcher]--
	}
	return true
}
```

PeterYurkovich: This can be deferred to the next PR though, I'm going to merge this one

PR #1177, opened 2026-08-24
COO-2185: refactor: remove i18n script in favor of env var configuration

jgbernalp:


jgbernalp:
/label qe-approved
to be checked from konflux FBCs

PR #1185, opened 2026-08-25
OU-1344: Add granular permission checks

PeterYurkovich:
This PR looks to update the patternfly UI to use granular permission checks for each action attempting to be taken. It uses a new `usePersesDashboardAccess` to check permissions for the user to perform a verb within a namespace. It entirely removes the `api/v1/permissions` call to Perses. 

All permission checks are delayed until they must be performed. This means that any user can open the create button now, even though they may not be able to create a dashboard in any namespace. Instead the error message within the dialog will let them know they do not have permission for the action they are trying to take in the namespace they have selected. The item dropdown is now no longer fully disabled, as the clone action requires the user to select a namespace, and only then will the denial permission check be shown.

This is the first of a 3 part approach which can be viewed in the connected jira. While this implementation has some more jagged edges, it fixes the immediate issue of the permission api call timing out and enables users to use the UI even if they are met with permission error messages more than before

## Summary by CodeRabbit

- **New Features**
  - Added project-specific access checks for creating, importing, editing, duplicating, renaming, and deleting dashboards.
  - Dashboard workflows now provide clear access-denied guidance and appropriately enable or disable actions.
  - Dashboard navigation now uses the shared namespace selector.

- **Bug Fixes**
  - Viewable projects remain visible even when users cannot modify them.
  - Restricted actions are consistently denied or disabled across dashboard workflows.
  - Improved duplicate-name validation messaging.


web/src/features/perses-dashboards/hooks/usePersesDashboardAccess.ts
```diff
@@ -0,0 +1,15 @@
...
+  verb: DashboardVerb,
+  namespace: string | null = null,
+): [boolean, boolean] => {
+  return useAccessReview({
+    group: 'perses.dev',
+    resource: 'persesdashboards',
+    verb,
+    namespace,
+  });
+};
```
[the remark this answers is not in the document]
PeterYurkovich: useAccessCheck requires a cluster to query against, any unit test would just be validating strings get passed through functions and would be useless

web/src/features/perses-dashboards/components/DashboardCreateDialog.tsx
```diff
@@ -77,14 +68,19 @@ export const DashboardCreateDialog: FC<DashboardCreateDialogProps> = ({ isOpen,
     },
   });
 
+  // eslint-disable-next-line react-hooks/incompatible-library
+  const selectedProject = form.watch('projectName');
+  const [canCreate, checkingAccess] = usePersesDashboardAccess('create', selectedProject || null);
+  const createDenied = !!selectedProject && !checkingAccess && !canCreate;
```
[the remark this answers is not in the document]
PeterYurkovich: Tested through e2e test updates. No update needed

web/src/features/perses-dashboards/pages/dashboard-list-page/DashboardListFrame.tsx
```diff
@@ -1,17 +1,26 @@
...
+  const { perspective } = usePerspective();
+  const navigate = useNavigate();
+
   return (
     <>
-      <ProjectBar activeProject={activeProject} />
+      <NamespaceBar
+        onNamespaceChange={() => {
+          const url = `${getDashboardsListUrl(perspective)}`;
+          navigate(url);
```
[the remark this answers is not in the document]
PeterYurkovich: This page isn't available in the dev perspective, so migrating to the current page is fine

web/src/features/perses-dashboards/components/DashboardDialogHelpers.tsx
```diff
@@ -1,61 +1,75 @@
...
-    }
-    return editableProjects.map((project) => ({
+    return availableProjects.map((project) => ({
       name: project,
       value: project,
       content: project,
       children: project,
     }));
-  }, [editableProjects]);
+  }, [availableProjects]);
```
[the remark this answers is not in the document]
PeterYurkovich: Functionality is tested by e2e tests, not needed here

web/src/features/perses-dashboards/pages/dashboard-page/DashboardToolbar.tsx
```diff
@@ -38,59 +38,31 @@ export interface DashboardToolbarProps {
...
       sx={{ whiteSpace: 'nowrap', minWidth: 'auto' }}
       data-test={persesDashboardDataTestIDs.editDashboardButtonToolbar}
     >
-      {loading ? t('Loading...') : t('Edit')}
+      {updateLoading ? t('Loading...') : t('Edit')}
     </Button>
   );
 
-  if (disabled && !loading) {
+  if (!updateLoading && !updateLoading) {
```
jgbernalp:
```suggestion
  if (!canUpdate && !updateLoading) {
```

web/src/features/perses-dashboards/hooks/usePersesDashboardAccess.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const usePersesDashboardAccess = (
+  verb: DashboardVerb,
+  namespace: string | null = null,
+  enabled = true,
+): [boolean, boolean] => {
+  // set to {} when not enabled to prevent fetching access
+  const resourceAttributes: AccessReviewResourceAttributes = enabled
+    ? { group: 'perses.dev', resource: 'persesdashboards', verb, namespace }
+    : {};
```
jgbernalp: when enabled is false, would this generate a permission error, for users that have access, while the request is processing? Basically a flicker

web/src/features/perses-dashboards/pages/dashboard-list-page/DashboardList.tsx
```diff
@@ -54,26 +52,28 @@ const DashboardActionsCell = memo(
...
     onDuplicate: (dashboard: DashboardResource) => void;
     onDelete: (dashboard: DashboardResource) => void;
-    emptyActions: { title: string; onClick: () => void }[];
   }) => {
     const { t } = useTranslation(process.env.I18N_NAMESPACE);
 
-    const { permissionsLoading } = useEditableProjects();
-    const { canEdit } = usePersesEditPermissions(project);
-    const disabled = !canEdit;
+    const [canUpdate, updateChecking] = usePersesDashboardAccess('update', project);
```
jgbernalp: would this generate an access review request per row render? or is this component mounted only when the actions are expanded?

web/src/features/perses-dashboards/hooks/usePersesDashboardAccess.ts
PeterYurkovich: Yes, that is why each component handles the loading as well. We could change the default behavior to be true, but I think it's better to handle the access and loading in a custom way in each component, so that they can pass items into their different components. For example the `EditButton` in `DashboardToolbar.tsx` will show `Loading...` instead of allowed showing the button as enabled or disabled while the loading is occuring

web/src/features/perses-dashboards/pages/dashboard-page/DashboardToolbar.tsx
PeterYurkovich: Fixed

web/src/features/perses-dashboards/pages/dashboard-list-page/DashboardList.tsx
PeterYurkovich: Fixed

web/cypress/support/commands/coo-install-commands.ts
```diff
@@ -338,125 +338,48 @@ export const cooInstallUtils = {
...
-                cy.log(`${elapsed}ms - Force delete output: ${forceResult.stdout}`);
-                if (forceResult.code !== 0) {
-                  cy.log(
-                    `Force delete failed with exit code ${forceResult.code}: ${forceResult.stderr}`,
-                  );
+                  Cypress.log({
+                    name: 'cleanupCOONamespace',
+                    message: `${MCP.namespace} is successfully deleted.`,
+                  });
+                  return cy.wrap(true, { log: false });
```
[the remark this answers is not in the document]
PeterYurkovich: This is a fine assumption to make and won't be fixed

PeterYurkovich:
I'm pretty sure the failures are due to running the ACM tests. Should be resolved when #1221 merges and this is rebased on top of it

PeterYurkovich:
Perses tests are failing due to package upgrade, not a fault of this PR

PeterYurkovich:
/hold
Waiting for we approved to push

PR #1202, opened 2026-08-31
OCPBUGS-101991: CVE-2026-33814 openshift4/ose-monitoring-plugin-rhel9: Go HTTP/2: Denial of Service via malformed SETTINGS_MAX_FRAME_SIZE frame [openshift-4.22]

tonyxrmdavidson:
Remediates CVE-2026-33814, a denial-of-service vulnerability in Go's HTTP/2 implementation (golang.org/x/net) where a malformed SETTINGS_MAX_FRAME_SIZE frame can be used to crash or hang a server.

Changes

- Added a replace directive in go.mod pointing golang.org/x/net to the patched fork github.com/openshift-sustaining/net@v0.50.0-sec.4.
- Bumped golang.org/x/net from v0.34.0 → v0.49.0 (via the replace) along with transitively affected golang.org/x/sys, golang.org/x/term, and golang.org/x/text to satisfy go mod tidy.
- Updated go.sum accordingly.

Testing

- make build-backend — builds successfully with go 1.24.0.
- make test-backend — all tests in ./pkg/... pass.


jgbernalp:
@tonyxrmdavidson not sure this is the right approach to fix this, I think it will be better to upgrade the go version in `go.mod` to 1.26.6 which patches a series of other CVEs. For this, we can check with the ART team, which productizes this plugin, to update to an appropriate go base image for the version mentioned before. Also upgrade to `golang.org/x/net v0.56.0` rather than adding a replace directive that will be harder to maintain. WDYT?

germanparente:
@jgbernalp in sustaining we fix the golang cves in all golang versions. So as not to have to bump golang. In this case 4.22 is built by default by ART with 1.25 and we try to keep this fixed.

PR #1216, opened 2026-09-01
OCPBUGS-103703: CVE-2026-33814-4-19 openshift4/ose-monitoring-plugin-rhel9: Go HTTP/2: Denial of Service via malformed SETTINGS_MAX_FRAME_SIZE frame [openshift-4.19]

tonyxrmdavidson:
golang.org/x/net's HTTP/2 implementation is vulnerable to a denial-of-service attack (CVE-2026-33814). A malicious or malformed peer can send an HTTP/2 SETTINGS frame with an invalid/out-of-range SETTINGS_MAX_FRAME_SIZE value. The affected code fails to properly validate this setting before using it to size internal buffers/frame handling, allowing a remote peer to trigger excessive resource consumption or a crash in the server, resulting in denial of service.

Fix: added a replace directive in go.mod (replace golang.org/x/net => github.com/openshift-sustaining/net v0.50.0-sec.4) swapping upstream golang.org/x/net for a patched fork that backports the upstream fix for CVE-2026-33814. The require block still lists golang.org/x/net v0.49.0 // indirect; this is expected, as replace unconditionally overrides that at build time (confirmed via go list -m golang.org/x/net, which resolves to the patched fork). go.sum updated with corresponding checksums.

Testing: make build-backend builds successfully against the patched module; make test-backend passes all tests in ./pkg/....

Backend-only change (Go module dependency), no application code modified. Target branch: release-4.19.


PeterYurkovich:
`/lgtm`

PeterYurkovich:
I'll let gabriel finish his discussion on the 4.22 PR before merging this one

PR #1228, opened 2026-09-01
OU-1472: perses version upgrade fixes

PeterYurkovich:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

## Summary by CodeRabbit

* **Tests**
  * Improved dashboard editor test reliability when setting static variable values.
  * Updated query type selection checks to target the correct control.
  * Improved chart value assertions for both data-present and no-data states.


etmurasaki:
This e2e-perses will probably fail due to mismatch between perses upgrade and UI

PR #1257, opened 2026-09-10
OU-1220: Add column to display dashboard id in perses dashboard list page

zhuje:
**JIRA** 
- https://redhat.atlassian.net/browse/OU-1220 

**Image** 
- quay.io/jezhu/monitoring-plugin:u1220-perses-dashboard-id-sep10-2

**Screenshot** 

Screenshot 2026-09-10 at 4 03 05 PM

## Summary by CodeRabbit

- **New Features**
  - Dashboard lists now show each dashboard’s immutable identifier in a separate sortable **Name** column.
  - Added explanatory tooltip text for the dashboard identifier.
  - The previous dashboard name column is now labeled **Display Name** for clearer distinction.
  - Dashboard rows can be sorted by name, identifier, project, creation date, modification date, or tag count.


PeterYurkovich:
Like 90% of the time this field is going to be very close or the same as the actual name. Could we add something like the console has on other pages where the user can select the visible columns, that way this can be hidden by default? I also believe the bug in the console has been fixed that made us use our own data table so we can swap to using theirs now if you don't want to add the implementation on top of our custom one. 

If you want I can do that fix first so this PR can just use the console SDK datatable to add the hidden columns thing, just lmk

zhuje:
> Like 90% of the time this field is going to be very close or the same as the actual name. Could we add something like the console has on other pages where the user can select the visible columns, that way this can be hidden by default? I also believe the bug in the console has been fixed that made us use our own data table so we can swap to using theirs now if you don't want to add the implementation on top of our custom one. If you want I can do that fix first so this PR can just use the console SDK datatable to add the hidden columns thing, just lmk

Meaning, like swapping to use the console SDK's `<VirtualizedTable />`? I think the console is trying to [deprecate](https://github.com/openshift/console/blob/15c788bd1d58dfacacb4b6188363b442b9b03a9a/frontend/packages/console-dynamic-plugin-sdk/src/api/core-api.ts#L106-L140) it. If that's the case, we could build the 'Manage columns' feature into the existing data table (Patternfly [example](https://www.patternfly.org/extensions/component-groups/column-management-modal/)).

I don't want to add more work to your plate; I can add the hidden dashboard ID column, but I'll probably need to update the ticket scope /estimate and move it to the next sprint.

PeterYurkovich:
The console team had told me in april they were working towards distributing it in the SDK, but it looks like that isn't done yet https://redhat.atlassian.net/browse/CONSOLE-5131

It also looks like patternfly doesn't actually support this behavior already like I had expected it would have, so we would need to create a new modal, hook, tests, ect. 

Lets not worry about it and whenever the SDK gets updated to include the ConsoleDataView we can swap to it and take add the behavior then

web/src/features/perses-dashboards/pages/dashboard-list-page/DashboardList.tsx
PeterYurkovich, line 159: Could we pull this function out into a separate file similar to `filter-silences.ts` and add some unit tests like `filter-silences.spec.ts`

PR #1263, opened 2026-09-14
OU-1344: lazy load perses permissions

PeterYurkovich:


PeterYurkovich:
/label qe-approved
`release-coo-1.5-ocp-4.19` has not diverged yet from `release-coo-ocp-4.19` which has already had QE performed on it

PR #1265, opened 2026-09-15
OCPBUGS-114748:[release-5.0] DOMPurify: Cross-Site Scripting via IN_PLACE sanitization

Pranavgarud06:
Updated the dependency resolution for _CVE-2026-75838_ by bumping `dompurify` to the patched version `3.4.13`.
This change ensures the dependency is resolved to a version containing the upstream fix for the reported vulnerability.


web/package.json
```diff
@@ -166,7 +166,7 @@
   "overrides": {
     "echarts": "^5.6.0",
     "qs": "^6.14.1",
-    "dompurify": "^3.4.11",
+    "dompurify": "^3.4.13",
```
jgbernalp: The resolved version is different, we are moving to use pinned versions:

```suggestion
    "dompurify": "3.4.15",
```

